PR #150, opened 2021-10-26
Log full JSON result when adding a team

tspascoal:
When a team was added to a repo It was showing the result as [object object]

This dumps the full result as JSON on the logs


decyjphr (approved):
LGTM

PR #152, opened 2021-11-12
Add compliance report function

decyjphr:
## Overview
When safe-settings has been set up, it is helpful to initially see which repositories are not complying with the policies and conduct some impact analysis. 
In the future, we could implement a functionality where where safe-settings is run in a passive mode where it will only monitor and generate reports.

## How it would work
Going to <deployment URL>/admin/report will create a HTML report

## Caveat
The report is pretty basic


martinm82:
@decyjphr any chance this PR could be resurrected? I think this is a very valuable feature.

decyjphr:
Yes, definitely. I am on vacation till the 10th and when I am back, will resurrect the work that was started and paused.

PR #182, opened 2022-02-23
Don't synch repo if it is on the exclusion list

### LINKED ISSUE #181: Excluded repos are synched with settings when a repo is edited
## Problem Description

When a repository that is on the exclusion list is edited changes are applied to it.

Repro steps:

- add a `settings.yml` file
- add a branch protection rule to `admin` repo
- Check that the changes in `settings.yml` have been applied

When a change is made to a repository that triggers a `branch_protection_rule` , `repository.edited` or  `repository.created` event the `settings.yml` are applied to the repo even if the repo is on the excluded list.

This happens even the exclusion list is customized or the built in one is used. For example with the OOTB config this happens to `admin` repo

Settings are not applied to a repository that is excluded if changes occur

```

```

github.com

safe settings version 0.1.0-rc.26
probot version 12.2.0

tspascoal:
If a repository is configured on the exclusion list, then settings are not applied when `synch` is called.

Closes: #181

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -16,6 +16,9 @@ class Settings {
     const { payload } = context
     const settings = new Settings(nop, context, repo, config, ref)
     await settings.loadConfigs()
+    if(settings.isRestricted(repo.repo)) {
+      return;
+    }
     await settings.updateRepos(repo)
     await settings.handleResults()
   }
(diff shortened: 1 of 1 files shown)


decyjphr (approved):
Thanks for adding this fix.

PR #184, opened 2022-02-24
Updating docs

joshjohanning:
Some parts of the doc are out of date. Updating them based on my experience.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -162,10 +162,11 @@ Every deployment will need an [App](https://developer.github.com/apps/).
 #### Repository Permissions
 
 - Administration: **Read & Write**
-- Contents: **Read only**
+- Checks: **Read & Write**
+- Commit statuses: **Read & Write**
+- Contents: **Read & Write**
 - Issues: **Read & Write**
-- Single file: **Read & Write**
-  - Path: `.github/settings.yml`
+- Pull requests: **Read & Write**
 
 #### Organization Permissions
 
@@ -174,7 +175,11 @@ Every deployment will need an [App](https://developer.github.com/apps/).
 
 ### Events
 
+- Branch protection rule
+- Check run
+- Check suite
 - Push
+- Pull request
 - Repository
 
 1. Download the private key from the app.
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -20,7 +20,7 @@ You will first need to clone the source code to your local environment that will
 
 - If using the manifest flow, follow the steps [here](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app-from-a-manifest)
 
-- If not using the `manifest flow` then follow the steps in [Create the GitHub App](#Create the GitHub App)
+- If not using the `manifest flow` then follow the steps in [Create the GitHub App](#create-the-github-app)
 
 - Create `.env` from `.env.example`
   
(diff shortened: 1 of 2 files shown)


moensch:
@joshjohanning Thank you for this - we've just been debugging much of the same. My settings match yours now and almost everything works. Except for one thing.

On the **first** commit on a PR, the check is forever stuck in "initializing" (screenshot below). On the second commit to the branch (can even be an empty commit), the check runs successfully. I wonder if you're seeing the same?

image

docs/deploy.md
```diff
@@ -162,10 +165,11 @@ Every deployment will need an [App](https://developer.github.com/apps/).
 #### Repository Permissions
 
 - Administration: **Read & Write**
-- Contents: **Read only**
+- Checks: **Read & Write**
+- Commit statuses: **Read & Write**
+- Contents: **Read & Write**
```
moensch, line 170: Do you know why this is needed? Does this have to do with `auto_init` stuffs?

joshjohanning:
> On the **first** commit on a PR, the check is forever stuck in "initializing" (screenshot below). On the second commit to the branch (can even be an empty commit), the check runs successfully. I wonder if you're seeing the same?

Ahh, yep, I didn't even try a PR in my repo yet, but I can confirm I see the same. Looking at the logs in the console, it looks like it's trying to do a compare, but it's using an invalid commit hash as the base. See: 

```
status: 404
request: {
    method: "GET"
    url: "https://api.github.com/repos/joshjohanning-org-safe-settings/admin/compare/0000000000000000000000000000000000000000...5f7d787b1a886fd614a9a5b271a5c149a904750f"
    headers: {
      "accept": "application/vnd.github.v3+json",
      "user-agent": "probot/12.2.0 octokit-core.js/3.5.1 Node.js/14.19.0 (linux; x64)",
      "authorization": "token [REDACTED]"
    }
    request: {}
}
```

Not sure where in the code this is, but we should be doing: 

```
HEAD...5f7d787b1a886fd614a9a5b271a5c149a904750f
```

instead of 

```
0000000000000000000000000000000000000000...5f7d787b1a886fd614a9a5b271a5c149a904750f
```

Creating https://github.com/github/safe-settings/issues/185 to track

docs/deploy.md
joshjohanning: @moensch The `Contents` of Read and Write is needed to be able to rename the default branch based on what are in the settings, the `Checks` is for the `nop` check on the PR I think,  and not entirely sure on `Commit statuses`  - I think having to deal with the [checks](https://docs.github.com/en/rest/reference/commits#list-commit-statuses-for-a-reference) again

decyjphr:
We have seen this weird behavior with commits becoming 00000... I believe this is an issue that is caused when sfe-settings is running in `heroku` and heroku for some reason doesn't like some branch names. Not sure if that is the case here.

joshjohanning:
> We have seen this weird behavior with commits becoming 00000... I believe this is an issue that is caused when sfe-settings is running in `heroku` and heroku for some reason doesn't like some branch names. Not sure if that is the case here.

Was not running in heroku for my instance - was running just on an Azure VM.

tspascoal:
@decyjphr  I've seen the same error as well (was running locally, not on heroku )

svg153:
Added new information in https://github.com/github/safe-settings/issues/185#issuecomment-1074086423

PR #188, opened 2022-03-03
Only load configs if not previously fetched on the same operation

tspascoal:
loading configs (repos,suborgs) was being loaded even when they had been loaded in the same instance.

This meant that on a synch all, configs were being loaded for each repo.

On an org with a moderate number of repos the decrease in calls is 10x

In some situations this would also trigger abuse limiting measures on GitHub


decyjphr (approved):
LGTM

PR #189, opened 2022-03-03
Don't get repo twice when handling default branch protection

tspascoal:
When dealing with the default branch in branches plugin, the repo was read even though it had already been fetched.

This reduces the number of calls in a synch, it could be further optimized by only getting it if the default branch is reference. But not defining the default repo is highly likely an edge case.


decyjphr (approved):
👍🏾

PR #190, opened 2022-03-04
Only update topics if there are changes

tspascoal:
If there are no changes to the topics no need to update it


decyjphr (approved):
A good call to avoid unnecessary API requests 👍🏾

tspascoal:
@decyjphr I don't have permissions to merge :)

decyjphr:
@tspascoal I added you as collaborator with write permissions for the future. I also merged that PR.

PR #191, opened 2022-03-04
Support team visibility (team creation only)

tspascoal:
Allow team's visibility to be set (closed or secret) when a team is created

If the team already exists it doesn't change it's visibility, this only applies for teams that are created.


decyjphr (approved):
Thanks for adding this.

PR #194, opened 2022-03-08
Only process applicable repos for suborg setting change

decyjphr:
The previous code did not handle suborg setting change efficiently. The PR has the optimization to only process the repos that are applicable for a suborg change and skips processing other repos.

This will reduce the number of API calls being made for suborg setting changes,

The PR also includes some cleanup and NPM package updates.


lib/settings.js
```diff
@@ -439,7 +457,10 @@ ${this.results.reduce((x,y) => {
    */
   async getSubOrgConfigs() {
     try {
-      const overridePaths = await this.getSubOrgConfigMap()
+      if (this.subOrgConfigMap) {
+        this.log.debug(`SubOrg config was changed and the associated overridePaths is = ${JSON.stringify(this.subOrgConfigMap)}`)
+      }
+      const overridePaths = this.subOrgConfigMap || await this.getSubOrgConfigMap()
```
tspascoal, line 535: Doesn't this mean we still load all suborg configs? (even if one (or maybe more)) have changed?

I'm aware that by reading only the change suborg files, we may miss repos that are **also** defined in other suborgs, but that shouldn't be done since it leads to undeterministic 
results.

What is the cost of loading all suborgs with a decent number of suborgs?

index.js
```diff
@@ -307,6 +271,7 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
 
     let suborg = getModifiedSubOrgConfigName(payload)
```
tspascoal: What happens if more than one suborg is pushed on a commit?

Isn't only the first one synched?

index.js
```diff
@@ -188,37 +150,39 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
-        return ( s.search(repoSettingPattern)>=0 )
+        return ( s.search(settingPattern)>=0 )
       }) !== undefined )
     })
 
     if (commit) {
-      robot.log.debug(`${JSON.stringify(commit)} \n ${commit.modified[0].match(repoSettingPattern)[1]}`)
-      return repo = {suborg: commit.modified[0].match(repoSettingPattern)[1], org: payload.repository.owner.name}
+      const matches = commit.modified[0].match(settingPattern)
+      robot.log.debug(`${JSON.stringify(commit)} \n ${matches[1]}`)
```
tspascoal: If there is not match, then `matches[1]` will crash.

This may happen if you have in the same commit both a `settings.yml` and at least one suborg because find is getting any commit that contains one suborg (but it may contain other files as well that are not suborgs).

If the first file on that commit is not a suborg this will fail

index.js
```diff
@@ -470,7 +435,7 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     }
     const suborg = getChangedConfigName(new Glob(".github/suborgs/*.yml"), files, context.repo().owner)
     if (suborg) {
-      return syncAllSettings(true, context, suborg, pull_request.head.ref )
+      return syncSubOrgSettings(true, context, suborg, context.repo(), pull_request.head.ref )
```
tspascoal: shouldn't the 
```
return syncAllSettings(true, context, context.repo(), pull_request.head.ref )
```

be the first thing to be done if there are settings? I mean if there is a `settings.yml` we are going to to a synch all anyway, not need to evaluate suborgs (and also handle the case if someone pushes a suborg and a settings in the same push)

something like
```
    if (settingsModified) {
      robot.log.debug(`Changes in '${Settings.FILE_NAME}' detected, doing a full synch...`)
      return syncAllSettings(false, context)
    }
```

tspascoal: I'm talking about the end of this block (but can't put a comment on something that hasn't changed).

But this is getting associated to the change from ` syncAllSettings`

This is confusing :)

tspascoal:
The logic  on the event handler for `push` will fail to do the right thing if multiple changes are done on a single commit and we may lose needed synchronizations

eg (some of the failing conditions):

- `settings.yml` + one suborg
- repo config + suborg (only repo is handled)
- Multi repo configs (only first one is handled)
- multiple suborgs (only first one is handled)

tspascoal (changes requested):

lib/settings.js
decyjphr: This is handled in the constructor which presets the `subOrgConfigMap` with the value of the modified suborg config file. So not all suborg configs are processed

index.js
decyjphr: This is still an unresolved issue. I have taken a technical debt to figure this out later when we have a good handle on the scale issue.

index.js
decyjphr: Yes, I agree. Moved that logic up.

lib/settings.js
```diff
@@ -111,38 +117,45 @@ ${this.results.reduce((x,y) => {
...
     }
 
     this.log.debug(`Completing check run ${JSON.stringify(params)}`)
     await this.github.checks.update(params)
-
   }
 
-  async loadConfigs() {
+  async loadConfigs () {
     this.repoConfigs = await this.getRepoConfigs(this.github, this.repo, this.log)
```
tspascoal: When this method is called in `suncSubOrgs` it is still loading all repos configuration(s) when it could (optimely) only load the repos for the suborg

index.js
```diff
@@ -302,28 +269,27 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
     }
 
     suborg = getAddedSubOrgConfigName(payload)
     if (suborg) {
       return syncSubOrgSettings(false, context, suborg)
     }
-    
+
     if (!settingsModified) {
       robot.log.debug(`No changes in '${Settings.FILE_NAME}' detected, returning...`)
```
tspascoal:
```suggestion
      robot.log.debug(`No changes detected, returning...`)
```

the `if (!settingsModified) {` and closing bracket should also go away. But can´t make a suggestion due to a deleted line

index.js
tspascoal: Since we potentially have situations on which changes are not acted on (eg: when there is more than one file changed (more than 1 repo config or more there one suborg (or combinations)) should we (for now) do an all synch if `settings.yml` or more than 2 files (other than `settings.yml`) are changed?

With a documented caveated that for now is recommended people only change one file per push/PR ?

index.js
decyjphr: I am looking at the logic to identify if more than one file has been changed and I feel it might be possible to use the same logic to call syncSettings or syncSubOrgSettings iteratively for the changed files.

tspascoal: That would be even better than the change I've suggested to synch all (it would be temporary until multiple files changes are supported).

index.js
decyjphr: @tspascoal I made changes to safe-settings where it will handle multiple files in the commit:
If it is settings.yml it will do syncall, otherwise it will do sync for the repos and  suborgs iteratively. 

Also, changes will be done only if there are actual changes based on `DeepCompare`. This makes it much more scalable. So API is called only when `DeepCompare` finds changes.

I did some benchmarking 1000 repos, with those modifications, I am seeing the following performance improvements:
1. Trigger sync all with changes to settings.yml but there are no changes in the settings: 2 minutes to fetch the current state for 1000 repos and compare the settings to determine there are no changes.
2. Trigger sync all with changes to settings.yml to add repo topics to all 1000 repos:  15 minutes
3. Changes to settings.yml to add branch protections to all the 1000 repos: 20 minutes
4. Run the PR check and perform validations for 1000 repos based on a change to settings.yml - 5 minutes

Probot automatically handles rate and abuse limiting. These numbers are including the delays to handle abuse limits and rate limits.  

I had to make changes to the way PR checks were reporting the status. Previously I was creating a detailed report which fails when completing the `check` because the payload size was too large. So I changed it to report just the summary.

index.js
```diff
@@ -455,33 +385,43 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+    const settingsModified = files.includes(Settings.FILE_NAME)
+
+    if (settingsModified) {
+      robot.log.debug(`Changes in '${Settings.FILE_NAME}' detected, doing a full synch...`)
+      return syncAllSettings(true, context, context.repo(), pull_request.head.ref)
+    }
+
+    const repoChanges = getChangedRepoConfigName(new Glob('.github/repos/*.yml'), files, context.repo().owner)
+    if (repoChanges.length > 0) {
+      return Promise.all(repoChanges.map(repo => {
```
tspascoal, line 412: Do we want to run this in parallel? It can trigger abuse limits if there are too many changes.....

index.js
```diff
@@ -455,33 +385,43 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
...
+        return syncSettings(true, context, repo, pull_request.head.ref)
+      }))
     }
-    const suborg = getChangedConfigName(new Glob(".github/suborgs/*.yml"), files, context.repo().owner)
-    if (suborg) {
-      return syncAllSettings(true, context, suborg, pull_request.head.ref )
+
+    const subOrgChanges = getChangedSubOrgConfigName(new Glob('.github/suborgs/*.yml'), files, context.repo().owner)
+    if (subOrgChanges.length) {
+      return Promise.all(subOrgChanges.map(suborg => {
```
tspascoal, line 419: Do we want to run this in parallel? It can trigger abuse limits if there are too many changes.....

tspascoal (approved):
LGTM

jetersen (approved):
Overall this looks like an improvement
Would love to see this merge 👍

Brecheyashtiff:
You know how about someone joining me in getting back at the people who are doing this to me. My wife killed herself and ever since everyone I know has bleed my dry , left me with NO ONE TO TRUST . I HAVE NO ONE.. I AM TO THE POINT OF SHOOTING MYSELF. SO THANKS FOR NOTHING. What or why do I care , I fell in love With Cheyenne. And she would rather see me dead and I guess that is what she wants , So Let It Be.. If you find my car they have stolen give it to my son... I loved my wife.. And I love my children .. And u love you Cheyanne. I just looked that shit up because I knew it would make you mad. And I knew about the spy app.  But I will kill myself now , or soon.. I am sorry I was never good enough

index.js
JAZZ1515: I got web3.Eth and other Eth networks. I have limited access to local host that's why I tried to fix this transaction where ever I felt comfortable.

There are several Pending Ether transactions from 12/08/22 I need to fix those as well but I was hoping help from Vitalik however he is busy. But I am new owner of Ethereum

index.js
JAZZ1515: no thanks. I think you guys have done a really good job. I appreciate

PR #202, opened 2022-03-17
Use time when creating and update checks

tspascoal:
The time was being set to midnight which showed odd values of checks in PRs


svg153:
Yes... it's confusing.
image

PR #203, opened 2022-03-25
check_suite.before returns null and check hangs

### LINKED ISSUE #185: The nop status check fails on pull requests with only 1 commit
## Problem Description

@moensch discovered that pull requests with only 1 commit fail the `nop` status check. 

This just "hangs" forever:

image

When you send another commit to this branch, then it succeeds:

image

```
status: 404
request: {
    method: "GET"
    url: "https://api.github.com/repos/joshjohanning-org-safe-settings/admin/compare/0000000000000000000000000000000000000000...5f7d787b1a886fd614a9a5b271a5c149a904750f"
    headers: {
      "accept": "application/vnd.github.v3+json",
      "user-agent": "probot/12.2.0 octokit-core.js/3.5.1 Node.js/14.19.0 (linux; x64)",
      "authorization": "token [REDACTED]"
    }
    request: {}
}
```

Not sure where in the code this is, but I think we should be doing something like: 

```
HEAD...5f7d787b1a886fd614a9a5b271a5c149a904750f
```

instead of 

```
0000000000000000000000000000000000000000...5f7d787b1a886fd614a9a5b271a5c149a904750f
```

Running my own

github.com

electricgull:
This fixes #185  
check_suite.before will return null (00000...) if there is only 1 commit. The check_suite function will hang because github returns a 404.

### DIFF
--- a/index.js
+++ b/index.js
@@ -460,6 +460,14 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
     }
     robot.log.debug(`Updating check run ${JSON.stringify(params)}`)
     let res = await context.octokit.checks.update(params)
+
+    // guarding against null value from upstream libary that is 
+    // causing a 404 and the check to stall
+    // from issue: https://github.com/github/safe-settings/issues/185#issuecomment-1075240374
+    if (check_suite.before === "0000000000000000000000000000000000000000") 
+    {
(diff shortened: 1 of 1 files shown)


decyjphr:
💯 thanks.

PR #212, opened 2022-04-27
Add support for autolinks in safe-settings

markjm:
Includes autolinks as a setting the app can handle and manage. If the `autolinks` field is not populated, this is a no-op. 

Docs about autolinks here https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources

Follows defined "diffable" pattern, so no change really.

Also moved removals of diffs before additions due to bugs with the previous ordering (if you reach max number of autolinks/labels/etc, then adding before removing will cause issues) - removing before adding will "make space" when needed. Not a common issue, but I ran int it while testing

### DIFF
--- a/lib/plugins/autolinks.js
+++ b/lib/plugins/autolinks.js
@@ -54,7 +54,7 @@ module.exports = class Autolinks extends Diffable {
   async remove({ id }) {
     const attrs = {
       ...this.repo,
-      id,
+      autolink_id: id,
     };
     if (this.nop) {
       return new NopCommand(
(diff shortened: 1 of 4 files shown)


svg153:
Just to link similar PRs:
- [Added autolinks PRs](https://github.com/probot/settings/pull/545) in [probot/settings](https://github.com/probot/settings/pull/547)

jetersen (approved):

lib/plugins/diffable.js
```diff
@@ -81,6 +81,12 @@ module.exports = class Diffable {
       return this.find().then(existingRecords => {
         const changes = []
 
+        existingRecords.forEach(x => {
```
nitrocode, line 84: Why is this function moving ?

lib/plugins/autolinks.js
```diff
@@ -0,0 +1,69 @@
+const Diffable = require('./diffable');
```
nitrocode, line 1: Should this have a test too to make sure it works in the future ?

daronmcintosh:
any word on when this'll be merged?

PR #213, opened 2022-04-29
Allow running validators against arbitrary config objects, not just ones with names

markjm:
Because the validation was happening during array merging, we could only validate entries in arrays. However, some plugins (like `repositories`) do not accept an array, and others (like `autolinks`) do not have a `name` property. So here I am hoisting the logic into `childPluginsList`

IMO, ideally we could run a validator on the entier plugin config object, but that would break backwards compatibility, so I have added "back" the logic to run on a per array entry basis. If we are OK breaking that compat, then we may want to remove the isArray special case

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -1,8 +1,7 @@
 const path = require('path')
-const fs = require('fs')
 const Glob = require('./glob')
 const NopCommand = require('./nopcommand')
-const { throws } = require('assert')
+
 class Settings {
 
   static async syncAll(nop, context, repo, config, ref) {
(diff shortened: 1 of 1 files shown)


decyjphr (approved):
Good catch.

decyjphr:
I'll test it and merge it.

markjm:
Hi @decyjphr - I saw you merged my other changes. Thanks! Any blockers on this last one?

decyjphr:
No blockers but this one has changes that I have not finished testing. Hope to get it done in a few days.

PR #218, opened 2022-06-02
Upgrade and caret probot dependency

markjm:
Specifically hoping for this commit, but I can not use this because this dep version is locked
https://github.com/probot/probot/commit/646b6a9276f1873f82d938020849193dbd2da89d


decyjphr (approved):
lgtm

PR #250, opened 2022-07-27
Respond to repository.created event instead of doing it ourselves

markjm:
This is functionally the same as before, but now we can support repositories created from other bots


markjm:
@decyjphr please lmk what you think. We have another bot managing repo creation (basically migrating from other scm systems). This allows us to apply policy on those repos ASAP

decyjphr (approved):
Ah! I like this approach. 🙇🏾

PR #292, opened 2022-09-17
Does fix bug inside of the Repository plugin related to updateRepo function ambiguity

nicacioliveira:
## What's the purpose of this PR?

Fixes a bug on the repository plugin when we need a security update

### Context

Inside of the repository class, we have an ambiguous return to `updaterepo` function that makes a bug when this function is called with `nop` variable true.

When the nop var is true, the function `updaterepo` returns an array of operations and when the nop var is false, the function returns one object with the repository data. Assuming that the expected response when the function is called is one array and the function update security expects one object with the repository data, we have a bug in this place.

---

Error caused by this ambiguity problem cited above:

```
/Users/nicaciooliveira/vtex/safe-settings/lib/plugins/repository.js:220
          this.log(`Enabling Dependabot alerts for owner: ${repoData.data.owner.login} and repo ${repoData.data.name}`)
                                                                     ^
TypeError: Cannot read properties of undefined (reading 'data')
    at Repository.updateSecurity (/Users/nicaciooliveira/vtex/safe-settings/lib/plugins/repository.js:220:70)
    at /Users/nicaciooliveira/vtex/safe-settings/lib/plugins/repository.js:104:25
```
Ambiguous points:

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ node_modules/
 private-key.pem
 .env
 *.pem
+
+.vscode
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -110,7 +110,7 @@ module.exports = class Repository {
               return this.updateSecurity(resp.data, resArray)
             }
 
-            return this.updateSecurity(await this.github.repos.update(this.settings).resArray)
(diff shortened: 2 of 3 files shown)


decyjphr (approved):
👍🏾

PR #294, opened 2022-09-21
Allow removing teams from override config

markjm:
Consider the below scenario: I want to have a global team (say engineering), but want to exclude this team specifically from one repo (`<repo>`). Right now there is no way to do this, because omitting `permission` from the override will cause it to be merged from the base config.

With this change, specifying null permissions will remove the team, instead of unsuccessfully trying to add the team with an unknown permission.

`.github/settings.yml`:
```
teams:
  - name: engineering
    permission: push
```

`.github/repos/<repo>.yml`:
```
teams:
  - name: engineering
    permission:
```

The above configuration will cause `engineering` to have write access to all teams except `<repo>`

### DIFF
--- a/lib/plugins/diffable.js
+++ b/lib/plugins/diffable.js
@@ -37,7 +37,7 @@ module.exports = class Diffable {
     // Filter out all empty entries (usually from repo override)
     for(const entry of filteredEntries) {
       for(const key of Object.keys(entry)) {
-        if(entry[key] === null || entry[key === undefined]) {
+        if(entry[key] === null || entry[key] === undefined) {
           delete entry[key]
         }
       }
--- a/lib/plugins/diffable.js
+++ b/lib/plugins/diffable.js
@@ -33,6 +33,17 @@ module.exports = class Diffable {
 
   filterEntries () {
     let filteredEntries = Array.from(this.entries)
+
+    // Filter out all empty entries (usually from repo override)
+    for(const entry of filteredEntries) {
+      for(const key of Object.keys(entry)) {
+        if(entry[key] === null || entry[key === undefined]) {
+          delete entry[key]
+        }
+      }
+    }
+    filteredEntries = filteredEntries.filter(entry => Object.keys(entry).filter(key => !MergeDeep.NAME_FIELDS.includes(key)).length !== 0)
+
     // this.log.debug(` entries ${JSON.stringify(filteredEntries)}`)
     filteredEntries = filteredEntries.filter(attrs => {
       if (Array.isArray(attrs.exclude)) {
--- a/lib/plugins/diffable.js
+++ b/lib/plugins/diffable.js
@@ -34,16 +34,6 @@ module.exports = class Diffable {
   filterEntries () {
     let filteredEntries = Array.from(this.entries)
 
-    // Filter out all empty entries (usually from repo override)
-    for(const entry of filteredEntries) {
-      for(const key of Object.keys(entry)) {
-        if(entry[key] === null || entry[key] === undefined) {
-          delete entry[key]
-        }
-      }
-    }
-    filteredEntries = filteredEntries.filter(entry => Object.keys(entry).filter(key => !MergeDeep.NAME_FIELDS.includes(key)).length !== 0)
-
     // this.log.debug(` entries ${JSON.stringify(filteredEntries)}`)
     filteredEntries = filteredEntries.filter(attrs => {
       if (Array.isArray(attrs.exclude)) {
@@ -90,7 +80,7 @@ module.exports = class Diffable {
 
   sync () {
     if (this.entries) {
-      const filteredEntries = this.filterEntries()
+      let filteredEntries = this.filterEntries()
       // this.log.debug(`filtered entries are ${JSON.stringify(filteredEntries)}`)
       return this.find().then(existingRecords => {
         const mergeDeep = new MergeDeep(this.log, ignorableFields)
@@ -104,6 +94,16 @@ module.exports = class Diffable {
           }
           return Promise.resolve()
         }
+
+        // Filter out all empty entries (usually from repo override)
+        for (const entry of filteredEntries) {
+          for (const key of Object.keys(entry)) {
+            if (entry[key] === null || entry[key] === undefined) {
+              delete entry[key]
+            }
+          }
+        }
+        filteredEntries = filteredEntries.filter(entry => Object.keys(entry).filter(key => !MergeDeep.NAME_FIELDS.includes(key)).length !== 0)
 
         const changes = []
 
--- a/lib/mergeDeep.js
+++ b/lib/mergeDeep.js
@@ -1,4 +1,5 @@
 const mergeBy = require('./mergeArrayBy')
+const NAME_FIELDS = ['name', 'username']
 class MergeDeep {
   constructor (log, ignorableFields, configvalidators = {}, overridevalidators = {}) {
     this.log = log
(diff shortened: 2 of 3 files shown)


svg153:
Hi @markjm.

I was reviewing the change and I remembered that in the collaborators section of the settings.yml example file inside README.md, appears that repository can be excluded.

Why did you prefer to leave without permissions instead of excluding the respiratory in the settings file?

I think your way is better. But I mention it, to avoid having two different ways of doing the same thing. I don't know if excluding collaborators from repositories works though.

In any case, I think it would be nice to document it. Even if it's just a small line in the example file.

Thank you very much for your contributions.

decyjphr:
As @svg153 pointed out, this feature is already there:
```
- name: core
  permission: admin
  exclude:
    - actions-demo
 ```

This gives the ability to the team that manages org-level policies to control which teams could be added to which repos. 
On the other hand, your approach gives some leeway to the team to decide if they don't want a certain team to be added to their repos. I think both are valid use cases with the precedence typically being given to the org level setting. i.e. if a team has been excluded from a repo by the org level or sub-org level setting, it cannot be violated by a repo level setting.

**Comment by markjm** [comment]
It also makes sense to me to have both. I consider settings.yml to be the "admin" role, while repo-specific configs are for teams/owners. I don't think it makes sense for teams to need to alter global configs for their specific use-case (though do like that admins do have "override" capabilities, both from include/exclude) and validators.

I believe this approach continues to support previously-working include/exclude while adding the capability for specific repos. 

I think the change makes sense given we allow repos to make "additive" changes already (adding more teams), so it makes sense to support "negative" changes as well if we say repo.yml supports "overrides"

**Comment by markjm** [comment]
Any further concerns on this, or are we good to merge?

**Comment by decyjphr** [comment]
I was testing this and it works for the scenario above. But if I had `.github/settings.yml`:
```
teams:
  - name: engineering
    permission: push
```

And no repo setting for the repo for `engineering` team,  the team `engineering` is added correctly to the repo. However, if this was not the intention and I add a repo level setting like this:

`.github/repos/<repo>.yml`:
```
teams:
  - name: engineering
    permission:
```

It doesn't remove the team from the repo.

It is probably an issue in `compareDeep` and I want to fix that. Also want to update the README to illustrate this feature.

**Comment by markjm** [comment]
That is strange, as that is the exact scenario ive been testing. Just tested again and it is still working.

One thing I notice is the teams diffable expects you pass `slug`s as names (as evidenced where comparator operator is comparing config name with remote slug), but mergeDeep is only looking at `name` and `username`. im not sure if/how this might cause differing results, but seems like a possibility. For reference, we are using slugs as names and name != slug, so perhaps I am hitting a `hasChanges` case always. Perhaps in your case, the slugs are always === to the name, which causes different behaviour from mergeDeep

**Comment by decyjphr** [comment]
In my test, there is only one team in the settings.yml and repo.yml. So when that team entry is deleted, it looks like there are no changes to be applied for teams and the app does not delete the team from GH.

File: lib/plugins/diffable.js [hunk 1]
```diff
@@ -33,6 +33,17 @@ module.exports = class Diffable {
...
+    // Filter out all empty entries (usually from repo override)
+    for(const entry of filteredEntries) {
+      for(const key of Object.keys(entry)) {
+        if(entry[key] === null || entry[key] === undefined) {
+          delete entry[key]
+        }
+      }
+    }
+    filteredEntries = filteredEntries.filter(entry => Object.keys(entry).filter(key => !MergeDeep.NAME_FIELDS.includes(key)).length !== 0)
+
```
decyjphr: Remove this step from here and move it further down in the sync method. That will take care of deleting any diffable that only has name and no other attributes or other attributes are null.

decyjphr (changes requested):

lib/plugins/diffable.js
decyjphr: This seems to fix the issue
```javascript
// Base class to make it easy to check for changes to a list of items
//
//     class Thing extends Diffable {
//       find() {
//       }
//
//       comparator(existing, attrs) {
//       }
//
//       changed(existing, attrs) {
//       }
//
//       update(existing, attrs) {
//       }
//
//       add(attrs) {
//       }
//
//       remove(existing) {
//       }
//     }
const MergeDeep = require('../mergeDeep')
const NopCommand = require('../nopcommand')
const ignorableFields = ['id', 'node_id', 'default']
module.exports = class Diffable {
  constructor (nop, github, repo, entries, log) {
    this.github = github
    this.repo = repo
    this.entries = entries
    this.log = log
    this.nop = nop
  }

  filterEntries () {
    let filteredEntries = Array.from(this.entries)

    // this.log.debug(` entries ${JSON.stringify(filteredEntries)}`)
    filteredEntries = filteredEntries.filter(attrs => {
      if (Array.isArray(attrs.exclude)) {
        if (!attrs.exclude.includes(this.repo.repo)) {
          // this.log.debug(`returning not excluded entry = ${JSON.stringify(attrs)} for repo ${this.repo.repo}`)
          return true
        } else {
          // this.log.debug(`skipping excluded entry = ${JSON.stringify(attrs)} for repo ${this.repo.repo}`)
          return false
        }
      } else {
        // this.log.debug(`No excludes. Returning unfiltered entries = ${JSON.stringify(attrs)} for repo ${this.repo.repo}`)
        return true
      }
    })
    filteredEntries = filteredEntries.filter(attrs => {
      if (Array.isArray(attrs.include)) {
        if (attrs.include.includes(this.repo.repo)) {
          // this.log.debug(`returning included entry  = ${JSON.stringify(attrs)} for repo ${this.repo.repo}`)
          return true
        } else {
          // this.log.debug(`skipping not included entry = ${JSON.stringify(attrs)} for repo ${this.repo.repo}`)
          return false
        }
      } else {
        // this.log.debug(`No includes. Returning unfiltered entries = ${JSON.stringify(attrs)} for repo ${this.repo.repo}`)
        return true
      }
    })
    filteredEntries = filteredEntries.map(e => {
      const o = {}
      for (const key in e) {
        if (key !== 'include' && key !== 'exclude') {
          Object.assign(o, {
            [key]: e[key]
          })
        }
      }

      return o
    })
    return filteredEntries
  }

  sync () {
    if (this.entries) {
      let filteredEntries = this.filterEntries()
      // this.log.debug(`filtered entries are ${JSON.stringify(filteredEntries)}`)
      return this.find().then(existingRecords => {
        const mergeDeep = new MergeDeep(this.log, ignorableFields)
        const compare = mergeDeep.compareDeep(existingRecords, filteredEntries)
        const results = JSON.stringify(compare, null, 2)
        this.log.debug(`Results of comparing ${this.constructor.name} diffable target ${JSON.stringify(existingRecords)} with source ${JSON.stringify(filteredEntries)} is ${results}`)
        if (!compare.hasChanges) {
          this.log.debug(`There are no changes for ${this.constructor.name} for repo ${this.repo}. Skipping changes`)
          if (this.nop) {
            return Promise.resolve([(new NopCommand(this.constructor.name, this.repo, null, `${this.constructor.name} settings has ${compare.additions.length} additions and ${compare.modifications.length} modifications`))])
          }
          return Promise.resolve()
        }

        // Filter out all empty entries (usually from repo override)
        for (const entry of filteredEntries) {
          for (const key of Object.keys(entry)) {
            if (entry[key] === null || entry[key] === undefined) {
              delete entry[key]
            }
          }
        }
        filteredEntries = filteredEntries.filter(entry => Object.keys(entry).filter(key => !MergeDeep.NAME_FIELDS.includes(key)).length !== 0)

        const changes = []

        existingRecords.forEach(x => {
          if (!filteredEntries.find(y => this.comparator(x, y))) {
            changes.push(this.remove(x))
          }
        })

        filteredEntries.forEach(attrs => {
          const existing = existingRecords.find(record => {
            return this.comparator(record, attrs)
          })

          if (!existing) {
            changes.push(this.add(attrs))
          } else if (this.changed(existing, attrs)) {
            changes.push(this.update(existing, attrs))
          }
        })

        if (changes.length === 0) {
          if (this.nop) {
            return Promise.resolve([
              // {plugin: this.constructor.name, repo: this.repo, action: `No changes`},
            ])
          }
        }
        return Promise.all(changes)
      }).catch(e => {
        this.log.error(`error calling find for ${this.constructor.name} ${e} for repo: ${JSON.stringify(this.repo)} entries ${JSON.stringify(this.entries)}`)
      })
    }
  }
}

```

markjm: @decyjphr applied review comments

decyjphr (approved):

PR #295, opened 2022-09-22
Wrong 'topics' variable value in repository plugin

nicacioliveira:
When the repository plugin will process repository topics, we needs a array of topics and not a string of topics.

The error occurs in the moment of the Repository instanciation when the topics is passed to this.topics like a string.
This is not handled error and the GitHub check run maintains an eternal loading state.

Example:

```javascript

'this.topics.join is not a function'
Stack:
'TypeError: this.topics.join is not a function\n    at Repository.updatetopics (/Users/nicaciooliveira/vtex/safe-settings/lib/plugins/repository.js:198:59)\n    
at /Users/nicaciooliveira/vtex/safe-settings/lib/plugins/repository.js:117:30\n    
at async Promise.all (index 284)\n    
at async Function.syncSubOrgs (/Users/nicaciooliveira/vtex/safe-settings/lib/settings.js:20:5)\n   
at async Promise.all (index 0)\n    at async Promise.all (index 0)\n    at async middleware (/Users/nicaciooliveira/vtex/safe-settings/node_modules/@octokit/webhooks/dist-node/index.js:355:5)'
```

Debugger:
image

the function join not works with a string, and this validations on this code not works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ private-key.pem
 *.pem
 
 .vscode
+yarn.lock
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -47,7 +47,7 @@ module.exports = class Repository {
     this.installationId = installationId
     this.github = github
     this.settings = Object.assign({ mediaType: { previews: ['nebula-preview'] } }, settings, repo)
-    this.topics = this.settings.topics
+    this.topics = settings && settings.topics && typeof settings.topics === 'string' ? this.settings.topics.split(',').map(t => t.trim()) : []
     this.security = this.settings.security
     this.repo = repo
     this.log = log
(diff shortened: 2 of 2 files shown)


brunojdo:
Hi, @decyjphr, do you had the chance to take a look at this? We are willing to use Safe Settings in our organization (2K+ repos) and we want to also contribute to the project. Thanks in advance

decyjphr:
I am not sure if I understand it properly, apologies for that. But safe-settings expects topics to be sent as an array in the config file, as shown [here](https://github.com/github/safe-settings/tree/main-enterprise/docs/sample-settings)

I think the confusion might be because in the old version the Topics were sent as a comma-delimited string. But that was changed some time back.

decyjphr:
Thanks, @nicacioliveira I think you are trying to accommodate the situation when it is passed as a comma-delimited string. Were you able to test this, or should I test this locally?

nicacioliveira:
Hello, Thank you too @decyjphr. You can test this locally to check if it's valid in the current version?

With your [commit](https://github.com/github/safe-settings/commit/596f022f05e7faa4e5eee63748d1589d6b5adf66). I think it is fixed for now. But if you can test locally, I would appreciate it.

PR #301, opened 2022-09-30
Empty (but not null) entries should clear teams and collaborators

Chris-V:
This conflicts with my other PRs so I'm opening as a draft. Also this handles the case in diffable.js, but maybe `mergeDeep.compareDeep` should instead output removed entries directly?

### DIFF
--- a/lib/plugins/collaborators.js
+++ b/lib/plugins/collaborators.js
@@ -20,35 +20,28 @@ module.exports = class Collaborators extends Diffable {
       this.github.repos.listCollaborators({ repo: this.repo.repo, owner: this.repo.owner, affiliation: 'outside' }),
       this.github.repos.listInvitations({ repo: this.repo.repo, owner: this.repo.owner })])
       .then(res => {
-        const results0 = res[0].data.map(user => {
+        const mapCollaborator = user => {
           return {
-          // Force all usernames to lowercase to avoid comparison issues.
+            // Force all usernames to lowercase to avoid comparison issues.
             username: user.login.toLowerCase(),
             pendinginvite: false,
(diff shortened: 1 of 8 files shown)


lib/plugins/collaborators.js
```diff
@@ -102,26 +95,23 @@ module.exports = class Collaborators extends Diffable {
...
-      )
+      return this.github.repos.deleteInvitation(data)
     } else {
+      const data = Object.assign({ username: existing.username }, this.repo)
       if (this.nop) {
         return Promise.resolve([
-          new NopCommand(this.constructor.name, this.repo, this.github.repos.removeCollaborator(
-            Object.assign(this.repo, { username: existing.username })
-          ), 'Remove Collaborator')
+          new NopCommand(this.constructor.name, this.repo, this.github.repos.removeCollaborator(data),
```
martinm82, line 110: Be careful. This will remove the collaborators. See https://github.com/github/safe-settings/pull/342/files#diff-aadaae6765a1167f3187e283961920c01d5db933ff8a7908ff10554b5520e689R117.

lib/plugins/teams.js
```diff
@@ -28,19 +28,19 @@ module.exports = class Teams extends Diffable {
 
   add (attrs) {
     let existing = { team_id: 1 }
-    this.log(`Getting team with the parms ${JSON.stringify(attrs)}`)
+    this.log.debug(`Getting team with the parms ${JSON.stringify(attrs)}`)
```
martinm82, line 31:
```suggestion
    this.log.debug(`Getting team with the params ${JSON.stringify(attrs)}`)
```

PR #324, opened 2022-11-10
Refactor Dockerfiles to be non root

GregUK:
Updated the build definitions to remove running as root when building the container.

Change the settings to inject the environment variables rather than hard coding them into the compose files or passing as individual env vars on the CLI.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -58,7 +58,7 @@ Once you have configured the **GitHub App** and updated the source code, you sho
 - Change directory to inside the code base
   - `cd safe-settings/`
 - Build the container
-  - `sudo docker build -t safe-settings .`
+  - `docker build -t safe-settings .`
 - This process should complete successfully and you will then have a **Docker** container ready for deployment
 
 #### Run the Docker container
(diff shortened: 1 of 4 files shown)


decyjphr (approved):
Looks good to me. I'll test it locally and merge.

docs/deploy.md
```diff
@@ -58,32 +58,32 @@ Once you have configured the **GitHub App** and updated the source code, you sho
...
+  - `docker build -t safe-settings .`
 - This process should complete successfully and you will then have a **Docker** container ready for deployment
 
 #### Run the Docker container
 Once the container has been successfully built, you can deploy it and start utilizing the **GitHub App**.
 
 #### Start the container with docker-compose
 If you have docker-compose installed, you can simply start and stop the **Docker** container with:
-- `cd safe-settings/; docker-compose up -d`
+- `cd safe-settings/; docker-compose --env-file .env up -d`
```
martinmigasiewicz-tomtom, line 69: Instead passing the --env-file option you could as well mount the `.env` file into the running container:

In docker-compose.yml:
```
- ${PWD}/.env:/opt/safe-settings/.env
```

martinmigasiewicz-tomtom (approved):

docs/deploy.md
martinmigasiewicz-tomtom, line 69: Instead passing the --env-file option you could as well mount the `.env` file into the running container:

In docker-compose.yml:
```
- ${PWD}/.env:/opt/safe-settings/.env
```

martinmigasiewicz-tomtom (approved):

docs/deploy.md
GregUK: This is true, I had opted for the command line argument so that it was more explicit which configuration file is being passed to the container.  Either approach is valid.

PR #342, opened 2022-11-26
Fix a severe bug in the code when the settings change might run even in the PR (NOP) mode. 

### LINKED ISSUE #339: Settings applied from pull request
Given a pull request in the `admin` repository with org `settings.yml`, it is possible that settings get applied on the whole organization even though the changes where not yet merged to the main branch.

Changes from branches/pull requests should never ever be applied before merging to the default branch.

I am reporting this issue since this definitely happened on our side and caused a lot of problems. I was actually expecting that changes get only applied on repositories when the PR is merged to main. I suspect that this happened during a manual trigger of the `re-run` feature of the safe-settings validator.

Self-hosted

GitHub.com

2.0.17

decyjphr:
Fix a severe bug in the code when the settings change might run even in the Pull Request (NOP) mode. 

When safe-settings runs in NOP mode, and there are changes to teams settings, it removes the existing teams even when run in the Pull request. This is caused by a bug in the code at https://github.com/github/safe-settings/blob/5dd8e7987f6bf685a44ad857d0c0cc3092cd7002/lib/plugins/teams.js#L75 where 

This fixes #339

### DIFF
--- a/lib/plugins/collaborators.js
+++ b/lib/plugins/collaborators.js
@@ -114,7 +114,7 @@ module.exports = class Collaborators extends Diffable {
     } else {
       if (this.nop) {
         return Promise.resolve([
-          new NopCommand(this.constructor.name, this.repo, this.github.repos.removeCollaborator(
+          new NopCommand(this.constructor.name, this.repo, this.github.repos.removeCollaborator.endpoint(
             Object.assign(this.repo, { username: existing.username })
           ), 'Remove Collaborator')
         ])
--- a/.github/workflows/create-pre-release.yml
+++ b/.github/workflows/create-pre-release.yml
@@ -70,6 +70,7 @@ jobs:
           docker run --env APP_ID=${{ secrets.APP_ID }} --env PRIVATE_KEY=${{ secrets.PRIVATE_KEY }} --env WEBHOOK_SECRET=${{ secrets.WEBHOOK_SECRET }} -d -p 3000:3000 yadhav/safe-settings:main-enterprise
           sleep 5
           curl http://localhost:3000
+      - run: echo "${{ github.head_ref }}"
       - name: Tag a final release 
         id: prerelease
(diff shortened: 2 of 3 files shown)


lib/plugins/collaborators.js
```diff
@@ -114,7 +114,7 @@ module.exports = class Collaborators extends Diffable {
     } else {
       if (this.nop) {
         return Promise.resolve([
-          new NopCommand(this.constructor.name, this.repo, this.github.repos.removeCollaborator(
+          new NopCommand(this.constructor.name, this.repo, this.github.repos.removeCollaborator.endpoint(
```
martinm82, line 117: Out of curiosity: What is the method `endpoint` doing? I could not find it in the docs: https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/master/docs/repos/removeCollaborator.md.

decyjphr: It is a method in the request that returns the params and the url that would be called by that particular request method. There is an example [here](https://octokit.github.io/rest.js/v18#custom-requests) and the code is [here](https://github.com/octokit/request.js/blob/master/src/with-defaults.ts)

martinm82 (approved):

lib/plugins/collaborators.js
martinm82: Just a dumb idea: How about making `NopCommand` making responsible for calling `.endpoint` in case `endpoint` argument is not null? It would require to provide the endpoint arguments as a separate parameter but at least you would ensure that NopCommand will be the only one calling these methods.

At least I can imagine that this bug might be introduced in the future again.

martinm82:
@decyjphr are you going to merge the fix soon?

lib/plugins/collaborators.js
decyjphr: Good idea, I'll make this change separately since it will affect the code broadly and we will need to test it more thoroughly

Issue #318, opened 2022-10-27
allow_auto_merge, delete_branch_on_merge and other valid fields are marked as ignorableFields


torrescd:
allow_auto_merge, delete_branch_on_merge and other valid fields are marked as ignorableFields

https://github.com/github/safe-settings/blob/5dd8e7987f6bf685a44ad857d0c0cc3092cd7002/lib/plugins/repository.js#L31

PR #344, opened 2022-11-28
Remove allow_auto_merge from ignored fields

glyph-se:
I tried applying `allow_auto_merge` using safe-settings though nothing happened. I then issue https://github.com/github/safe-settings/issues/318 and tried to change the code myself.

Removing it from `ignoredFields` make the `allow_auto_merge` setting to be applied properly.

I have to admit that I don't understand how the code works, but this change solves our issue. It is also possible that more things should be removed from the ignored list.

### DIFF
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -28,7 +28,6 @@ const ignorableFields = [
   'temp_clone_token',
   'allow_merge_commit',
   'allow_rebase_merge',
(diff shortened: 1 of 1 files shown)


decyjphr:
@glyph-se thank you for raising this PR, it looks like we can remove a few more items from the `ignorable` fields list. I'll need to test these and so will take me a few more days to merge this.

PR #356, opened 2022-12-08
Decyjphr/release pr 295

decyjphr:
Reverse merge main and test PR-295

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ private-key.pem
 *.pem
 
 .vscode
+yarn.lock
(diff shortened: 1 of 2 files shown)


martinm82:
@decyjphr isn't `topics` a list of strings? Currently this change results that if `topics` is defined as a list, it will be set always to an empty list.

Maybe instead it should be:

```
this.topics = settings && settings.topics && typeof settings.topics === 'string' ? this.settings.topics.split(',').map(t => t.trim()) : Array.isArray(settings.topic) ? settings.topics : []
```
?

PR #361, opened 2022-12-13
Convert topics string passed by mistake to an array

decyjphr:
This is to fix an issue in #295

### DIFF
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -47,7 +47,14 @@ module.exports = class Repository {
     this.installationId = installationId
     this.github = github
     this.settings = Object.assign({ mediaType: { previews: ['nebula-preview'] } }, settings, repo)
-    this.topics = settings && settings.topics && typeof settings.topics === 'string' ? this.settings.topics.split(',').map(t => t.trim()) : []
+    
+    this.topics = settings.topics
+   
+    // Topics should be an array and this will convert any string passed by mistake to an array
+    if ( typeof this.topics === 'string' ) {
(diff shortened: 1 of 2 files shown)


decyjphr:
@martinm82 I am resetting the change back. Best to keep topics as an Array. If I change it to allow for strings, I might cause issues in other places. For e.g. org level settings might have it as an array and repo level might have it as string and that causes issues in comparing.
Thanks for catching this issue. 🙏🏾

lib/plugins/repository.js
```diff
@@ -47,7 +47,7 @@ module.exports = class Repository {
     this.installationId = installationId
     this.github = github
     this.settings = Object.assign({ mediaType: { previews: ['nebula-preview'] } }, settings, repo)
-    this.topics = settings && settings.topics && typeof settings.topics === 'string' ? this.settings.topics.split(',').map(t => t.trim()) : []
+    this.topics = this.settings.topics
```
martinm82, line 50: Should we ensure that `this.settings.topics` is really an array?

decyjphr: I am not sure how to handle it if it is not an array. If it is a string, I can convert it to an array, but I have to contend with the situation where if the org and suborg level settings had topics as arrays, DeepMerge would have basically tried to compare a string with array and would have overwritten the Org and suborg level topics with just this string. This means we have to handle this in DeepMerge also.

Since it was not clear how to handle this, I just reverted it back to the way it was before.

martinm82 (approved):

PR #366, opened 2022-12-28
Fix for the bug where all the repo override files are not read

viks-confluent:
Description:

safe-settings use get-content api to fetch all the repo permission override files. That api has a limit of 1000 and our org had way > 1000 repo files in the admin repo. This has lead to incorrect repo permissions.

GitHub Api docs: https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content

Fix is to use the github tree api(https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree) to read all the repos directory content.

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -424,12 +424,51 @@ ${this.results.reduce((x, y) => {
   async getRepoConfigMap () {
     try {
       this.log.debug(` In getRepoConfigMap ${JSON.stringify(this.repo)}`)
-      const repo = { owner: this.repo.owner, repo: 'admin' }
-      const CONFIG_PATH = '.github'
-      const params = Object.assign(repo, { path: path.posix.join(CONFIG_PATH, 'repos'), ref: this.ref })
+      // GitHub getContent api has a hard limit of returning 1000 entries without
+      // any pagination. They suggest to use Tree api.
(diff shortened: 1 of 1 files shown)


viks-confluent:
@decyjphr  Please take a look at this when you get some time. Thanks!

decyjphr (approved):
@viks-confluent 👍🏾

lib/settings.js
```diff
@@ -424,12 +424,51 @@ ${this.results.reduce((x, y) => {
...
+      const githubDirContent = githubDirectoryContentResponse.data
+      const repoDirInfo = githubDirContent.filter(dir => dir.name === 'repos')[0]
+      if (!repoDirInfo) {
+        this.log.debug('No repos directory in the admin/.github')
+        return []
+      }
+
+      // read the repo contents using tree
+      this.log.debug(`repos directory info ${JSON.stringify(repoDirInfo)}`)
+      const endpoint = `/repos/${this.repo.owner}/${repo.repo}/git/trees/${repoDirInfo.sha}`
```
decyjphr, line 451: One option would be to replace `github.request` with some syntactic sugar `github.git.getTree({owner, repo, tree_sha}) ` https://octokit.github.io/rest.js/v18#git-get-tree

PR #373, opened 2023-01-04
Fixes for compareDeep

### LINKED ISSUE #372: branches: can't remove restrictions
## Problem Description

When you remove a push restriction from a branch protection rule, the change can't be applied.

Test scenario A:

1. Create a branch protection rule with two users or teams in the push restrictions. Example:

```yaml
branches:
  - name: default
    protection:
      restrictions:
        apps: []
        users: []
        teams: 
          - developers
          - marketing
```

2. Remove one user or team and push to admin repo
3. The app fails to apply the change

Test scenario B:

1. Create a branch protection rule with one user or team in the push restrictions. Example:

```yaml
branches:
  - name: default
    protection:
      restrictions:
        apps: []
        users: []
        teams: 
          - developers
```

2. Manually add another user or team to the restriction using the GitHub UI
3. The app fails to apply the change (remove the manually added user/team) when trying to enforce the policy stated in the yaml

"Error applying branch protection" in the log, with a 422 response with the following error from GitHub.

"Branch protection applied successfully" in the log

```
"data":{"message":"Invalid request.\n\nNo subschema in \"anyOf\" matched.\nFor 'items', {} is not a string.\nNot all subschemas of \"allOf\" matched.\nFor 'anyOf/1', {\"apps\"=>[], \"users\"=>[], \"teams\"=>[\"developers\", {}]} is not a null."
```

My own

github.com

5.0.20-alpha.4

N/A

## Source

TL;DR; 

1. compareDeep mutates its inputs
2. compareDeep produce unexpected results for lists

The reason for this bug is that `mergeDeep.compareDeep` mutates the given input that later on is sent to the GitHub API.
In my example, the function adds an empty object to the array `branch.protection.restrictions.teams` on this line: https://github.com/github/safe-settings/blob/15518fe7f14facf82138844948ac6ad20330ecfb/lib/plugins/branches.js#L50

When I set a breakpoint on that line, the `branch.protection.restrictions.teams` array is `["developers"]`, but after the call the it has been changed to `["developers", {}]`. 

The return value from `compareDeep` also contains unexpected data. 
When the yaml only has the team "developers" and current repo settings has one or more extra users/teams, the following result is generated:

```
DEBUG (event): Result of compareDeep = {
  "additions": {},
  "modifications": {
    "restrictions": {
      "teams": [
        "developers",
        {
          "0": "d",
          "1": "e",
          "2": "v",
          "3": "e",
          "4": "l",
          "5": "o",
          "6": "p",
          "7": "e",
          "8": "r",
          "9": "s",
          "name": "developers"
        }
      ]
    }
  },
  "hasChanges": true
}
```

If there are two teams in the yaml, and one _or more_ extra in the rule on GitHub, the invalid payload will be `teams:["developers", "marketing", {}]`.

Also, the result of compareDeep is even more strange when having two teams in the yaml:

```
DEBUG (event): Result of compareDeep = {
  "additions": {},
  "modifications": {
    "restrictions": {
      "teams": [
        "developers",
        "marketing",
        {
          "0": "d",
          "1": "e",
          "2": "v",
          "3": "e",
          "4": "l",
          "5": "o",
          "6": "p",
          "7": "e",
          "8": "r",
          "9": "s",
          "name": "developers"
        },
        {
          "0": "d",
          "1": "e",
          "2": "v",
          "3": "e",
          "4": "l",
          "5": "o",
          "6": "p",
          "7": "e",
          "8": "r",
          "9": "s",
          "name": "developers"
        },
        {
          "0": "m",
          "1": "a",
          "2": "r",
          "3": "k",
          "4": "e",
          "5": "t",
          "6": "i",
          "7": "n",
          "8": "g",
          "name": "marketing"
        }
      ]
    }
  },
  "hasChanges": true
}
```

If you have three teams, the first and second teams will be repeated once more for that iteration. This is obviously caused by the recursive call to compareDeep.

anderssonjohan:
fix #372 

The chore commits in this PR was added to be able to run `npm test` locally since the linters complained without being able to run the test suite.

The two fix commits are changes that will fix the two bugs described in #372.

Running the added tests without applying the changes to `mergeDeep.js` confirms the bugs as described in the issue:
Screenshot 2023-01-04 at 20 50 11

`npm test` result before this PR:

Screenshot 2023-01-04 at 20 47 12

`npm test` result with this PR:

Screenshot 2023-01-04 at 21 51 10
(the two added tests are passing)


lib/mergeDeep.js
```diff
@@ -55,6 +55,10 @@ class MergeDeep {
       return this.compareEmptyTarget(t, s, additions, modifications)
     }
 
+    if (typeof s === 'string' && (t.name === s || t.username === s)) {
```
martinm82: How about doing this instead:

```suggestion
    if (typeof s === 'string' && (Object.keys(t).some(key => NAME_FIELDS.includes(key)))) {
```

At least you get rid of the duplication of the `name` and `username` fields.

anderssonjohan: @martinm82 This avoids looping through the characters in the given conditions:

```javascript
t = { name: "something" }
s = "something"
```

..where "something" is a team name for example, thus the reason why I compare the values.
But I think you are right, we don't need to do that here. We can use your suggestion to do the early exit no matter the value for the given object's name/username property. We only need to check the shape of the inputs. Thank you!

anderssonjohan: I just pushed a fixup commit for this, I'll rebase and squash it into the other commit relevant to this when I have approval on the PR.

martinm82: BTW, how about adding some comment explaining the why. Today I understand this block but tomorrow I for sure won't understand the reasoning behind it. Yes, there is git blame and I probably would find this pull request again.

anderssonjohan: Yeah, I thought the commit would explain it, but I can for sure add a comment in the code about it.

Screenshot 2023-01-09 at 09 57 15

anderssonjohan: @martinm82 Another thought.. Would it ever be reasonable to continue the comparison from here if the source `s` is a string? I mean, that would never add any useful info to the arrays (additions, modifications).

What I mean is that I can simplify it down to:

```javascript
if (typeof s === 'string') {
  return
}
```

martinm82: Let me think about it but in general if we have a good test coverage and test cases are passing as expected we might indeed skip the other condition.

anderssonjohan: @martinm82 I pushed a fixup commit now with a comment and the simplification that I mentioned above. The guard could also be placed in all locations of the recursive calls. Given the current design of the function, the recursive calls should never take place for primitives such as strings. (primitives are compared in the else block of the loop)

Another design is to recurse into the function even for primitives and move the comparison up from the else block in the loop and to this guard before the loop starts. That however would add more recursive calls but with the tradeoff that the function can be called with primitive values (such as two strings) and have the differences added to the given arrays (additions, modifications). It's however a bit weird IMHO to use the function like this with primitives. When comparing primitives it would probably make more sense with a conventional compare function that returns -1, 0, or 1.

anderssonjohan:
Rebased and squashed fixup commits, ready for review + merge at will.

martinm82 (approved):
LGTM

But since I am not the author of this I would prefer @decyjphr to review and approve these changes as well.

anderssonjohan:
@decyjphr please let me know if you need something from me to merge this 👍

decyjphr:
I reverse-merged main to this and tested. It looks good. Will go ahead and merge.

PR #394, opened 2023-02-02
feat: add scheduler handler

### LINKED ISSUE #313: [Question] Safe-settings cron with lambda via serverless
I have my safe settings lambda up and running (via serverless) and verified everything works fine. I want to run these settings periodically via lambda setup. As this is a lambda function, node-cron does not make sense.

NOTE: I tried creating a Cloudwatch event to trigger lambda function. However that is triggering the Serverless handler webhook which is not actually running the settings bot code.

So what is the suggested approach in running the lambda periodically to reconfigure settings for all the repos in the organization ? Any documentation would be appreciated.

anderssonjohan:
Provides example configuration for using EventBridge events to run `syncInstallation` on a schedule in a separate Lambda function.

- The provided example trigger is disabled by default
- Manual invocation is possible by clicking the Test button in the Test tab of the Lambda in the AWS Console

Implemented based on the example provided by @viks-confluent in https://github.com/github/safe-settings/issues/313#issuecomment-1295371756

fixes #313

### Screenshots

Output from manual run

Screenshot 2023-02-02 at 17 52 10

A useful query for CloudWatch Log Insights to get logs from both log groups (webhook invocations and scheduled/manual runs):
Screenshot 2023-02-02 at 17 53 12

### DIFF
--- a/handler.js
+++ b/handler.js
@@ -8,3 +8,9 @@ const appFn = require('./')
 module.exports.webhooks = createLambdaFunction(appFn, {
   probot: createProbot()
 })
+
+module.exports.scheduler = function () {
+  const probot = createProbot()
+  const app = appFn(probot)
+  return app.syncInstallation()
+}
--- a/index.js
+++ b/index.js
@@ -459,4 +459,8 @@ module.exports = (robot, _, Settings = require('./lib/settings')) => {
       syncInstallation()
     })
   }
+
+  return {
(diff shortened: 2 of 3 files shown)


anderssonjohan:
ping @decyjphr

decyjphr (approved):
Thanks @anderssonjohan This will be of great value to those using Lambda

PR #371, opened 2023-01-03
feat: support for regular expressions in restrictedRepos

martinm82:
In our org we would like to skip certain repositories (playgrounds, dev, PoC, etc.) from being configured with safe-settings. 

This PR introduces support for using regular expressions in the `restrictedRepos` attributes `include` and `exclude`.

PR #395, opened 2023-02-02
docs: use strict repo names in sample config

anderssonjohan:
Aligns the provided sample config with the feature added in github/safe-settings#371 so it continues to match the intended repositories.

@decyjphr @martinm82 The change in #371 had some surprising side effects for me. Any repo containing the word `admin` became excluded and I also had my test repository named `safe-settings-test` excluded by the string `'safe-settings'`.


martinm82:
Yes, I agree with this change. I hit this issue as well. Sorry that I missed that in the PR.

martinm82 (approved):

PR #399, opened 2023-02-07
feat: add the ability to enable Automated Security Fixes

amenocal:
There are two `security` options to enable for a repository

- [Enable Vulnerability Alerts](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#enable-vulnerability-alerts)
- [Enable Automated Security Fixes](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#enable-automated-security-fixes)

Currently only **Enable Vulnerability Alerts** is being enabled as part of safe-settings.

This PR will enable to decide which flag to enable/disable

### DIFF
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -104,7 +104,7 @@ module.exports = class Repository {
               return this.updateSecurity(resp.data, resArray)
             }))
           promises.push(this.updaterepo(resArray).then(() => {
-            return this.updateSecurityDependabotAlerts(resp.data, resArray)
+            return this.updateAutomatedSecurityFixes(resp.data, resArray)
           }))
         }
         if (topicChanges.hasChanges) {
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -100,12 +100,14 @@ module.exports = class Repository {
           }
           // Remove topics as it would be handled seperately
           delete this.settings.topics
-            promises.push(this.updaterepo(resArray).then(() => {
+          promises.push(this.updaterepo(resArray).then(() => {
               return this.updateSecurity(resp.data, resArray)
             }))
           promises.push(this.updaterepo(resArray).then(() => {
             return this.updateAutomatedSecurityFixes(resp.data, resArray)
           }))
+        } else {
+          promises.push(this.updateSecurity(resp.data, resArray))
         }
         if (topicChanges.hasChanges) {
           promises.push(this.updatetopics(resp.data, resArray))
@@ -224,7 +226,7 @@ module.exports = class Repository {
             resArray.push((new NopCommand(this.constructor.name, this.repo, this.github.repos.enableVulnerabilityAlerts.endpoint({
               owner: repoData.owner.login,
               repo: repoData.name
-            }), 'Update Topics')))
+            }), 'Enabling Dependabot alerts')))
             return Promise.resolve(resArray)
           }
           return this.github.repos.enableVulnerabilityAlerts({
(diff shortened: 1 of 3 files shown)


lib/plugins/repository.js
```diff
@@ -100,8 +100,11 @@ module.exports = class Repository {
           }
           // Remove topics as it would be handled seperately
           delete this.settings.topics
+            promises.push(this.updaterepo(resArray).then(() => {
```
martinm82: Please correct the indentation.

lib/plugins/repository.js
```diff
@@ -252,4 +256,39 @@ module.exports = class Repository {
       }
     }
   }
-}
+
+  updateAutomatedSecurityFixes(repoData, resArray) {
+    if (this.security) {
+      if (this.security?.enableAutomatedSecurityFixes === true || this.security?.enableAutomatedSecurityFixes === false) {
```
martinm82, line 261: This condition is superfluous since you already have for both of them a separate block below.

lib/plugins/repository.js
```diff
@@ -252,4 +256,39 @@ module.exports = class Repository {
...
+        if (this.security.enableAutomatedSecurityFixes === true) {
+          this.log(`Enabling Dependabot security updates for owner: ${repoData.owner.login} and repo ${repoData.name}`)
+          if (this.nop) {
+            resArray.push((new NopCommand(this.constructor.name, this.repo, this.github.repos.enableAutomatedSecurityFixes.endpoint({
+              owner: repoData.owner.login,
+              repo: repoData.name
+            }), 'Update Topics')))
```
martinm82: I guess this is a copy&paste error. This call has nothing to do with topics.

lib/plugins/repository.js
```diff
@@ -252,4 +256,39 @@ module.exports = class Repository {
...
+            repo: repoData.name
+          })
+        }
+        if (this.security.enableAutomatedSecurityFixes === false) {
+          this.log(`Disabling Dependabot security updates for owner: ${repoData.owner.login} and repo ${repoData.name}`)
+          if (this.nop) {
+            resArray.push((new NopCommand(this.constructor.name, this.github.repos.disableAutomatedSecurityFixes.endpoint({
+              owner: repoData.owner.login,
+              repo: repoData.name
+            }), 'Update Topics')))
```
martinm82: Same as above: copy&paste error

martinm82 (changes requested):

lib/plugins/repository.js
amenocal: @martinm82 agree 🤔 , as this is my first contribution, I figured I would follow the same style as the [updateSecurity](https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/repository.js#LL217C25-L217C25) , let me know if you'd like to remove it from both functions

lib/plugins/repository.js
amenocal: @martinm82 yeap, and it seems to be coming from `updateSecurity` as well. I'll go ahead and update both.

lib/plugins/repository.js
martinm82: If you want you can change it as well in the other places but I am not insisting ;)

amenocal: @martinm82 I've updated both functions 😄

lib/plugins/repository.js
```diff
@@ -235,21 +239,54 @@ module.exports = class Repository {
...
+              repo: repoData.name
+            }), 'Enabling Dependabot security updates')))
+            return Promise.resolve(resArray)
+          }
+          return this.github.repos.enableAutomatedSecurityFixes({
+            owner: repoData.owner.login,
+            repo: repoData.name
+          })
+        }
+        if (this.security.enableAutomatedSecurityFixes === false) {
```
martinm82: This could be a regular `else` block.

martinm82 (approved):

robandpdx (approved):

PR #408, opened 2023-02-10
feat(labels): support exclusion regexp

### LINKED ISSUE #407: Labels test fails due to change in requirement
The following filter:
https://github.com/github/safe-settings/blob/e2ffb51e4a2802cff13fa9ee6129f617bd150888/lib/plugins/diffable.js#L98

..was added back in https://github.com/github/safe-settings/commit/b3803cdecd685510fda735ff901bbe4b12af8d6e#diff-00b30edef56c2bcf34272d0571801eeffdb0937c92b977bc559a653f97c0d657R45

This makes the labels test fail where the setup uses the following label object:

```
{ name: 'added' }
```

Labels with only name attribute will be deleted because of this filter.
The tests fails due to `createLabel` not being called because the label with name `'added'` is filtered out.

anderssonjohan:
fixes #407
fixes https://github.com/github/safe-settings/discussions/398

Adds support for alternative `labels` configuration where labels matching a given pattern can be excluded from delete operations. This allows other tools to create labels without having them deleted by safe-settings.

Example: The following config would keep the labels `released`, `released on @dev`, `released on @7.x` created by a tool such as Semantic Release.

```
labels:
  include:
    - name: bug
      color: '#ff0000'
  exclude:
    - name: ^release
```

Include can be used without specifying exclude in the config, while also supporting the current format:

```
labels:
  - name: bug
    color: '#ff0000'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,24 +322,28 @@ repository:
 # The app will apply the following settings to it
 labels:
   # Labels: define labels for Issues and Pull Requests
-  - name: bug
-    color: CC0000
-    description: An issue with the system
+  include:
+    - name: bug
+      color: CC0000
+      description: An issue with the system
 
-  - name: feature
-    # If including a `#`, make sure to wrap it with quotes!
-    color: '#336699'
-    description: New functionality.
+    - name: feature
+      # If including a `#`, make sure to wrap it with quotes!
+      color: '#336699'
+      description: New functionality.
 
-  - name: first-timers-only
-    # include the old name to rename an existing label
-    oldname: Help Wanted
-    color: '#326699'
+    - name: first-timers-only
+      # include the old name to rename an existing label
+      oldname: Help Wanted
+      color: '#326699'
 
-  - name: new-label
-    # include the old name to rename an existing label
-    oldname: Help Wanted
-    color: '#326699'
+    - name: new-label
+      # include the old name to rename an existing label
+      oldname: Help Wanted
+      color: '#326699'
+  exclude:
+    # don't delete any labels created on GitHub that starts with "release"
+    - name: ^release
 
 milestones:
 # Milestones: define milestones for Issues and Pull Requests
--- a/lib/plugins/labels.js
+++ b/lib/plugins/labels.js
@@ -3,8 +3,21 @@ const NopCommand = require('../nopcommand')
 const previewHeaders = { accept: 'application/vnd.github.symmetra-preview+json' }
 
 module.exports = class Labels extends Diffable {
-  constructor (...args) {
-    super(...args)
+  constructor (nop, github, repo, entries, log) {
+    const { include, exclude = [] } = entries
+    if (Array.isArray(include)) {
+      // Config is object with include array
+      // labels:
+      //   include:
+      //     - name: label
+      super(nop, github, repo, include, log)
+    } else {
+      // Config is array
+      // labels:
+      //   - name: label
+      super(nop, github, repo, entries, log)
+    }
+    this.exclude = exclude.filter(item => item?.name).map(item => new RegExp(item.name))
 
     if (this.entries) {
       this.entries.forEach(label => {
@@ -67,6 +80,9 @@ module.exports = class Labels extends Diffable {
   }
 
   remove (existing) {
+    if (this.isExcluded(existing)) {
+      return Promise.resolve([])
+    }
     if (this.nop) {
       return Promise.resolve([
         new NopCommand(this.constructor.name, this.repo, this.github.issues.deleteLabel.endpoint(this.wrapAttrs({ name: existing.name })), 'Delete label')
@@ -77,5 +93,9 @@ module.exports = class Labels extends Diffable {
 
   wrapAttrs (attrs) {
     return Object.assign({}, attrs, this.repo, { headers: previewHeaders })
+  }
+
+  isExcluded (label) {
+    return this.exclude.some(rx => rx.test(label.name))
   }
 }
--- a/test/unit/lib/plugins/labels.test.js
+++ b/test/unit/lib/plugins/labels.test.js
@@ -98,5 +98,51 @@ describe('Labels', () => {
         expect(github.issues.createLabel).toHaveBeenCalledTimes(1)
       })
     })
+
+    it('does not delete excluded labels', () => {
+      github.paginate.mockReturnValueOnce(Promise.resolve([
+        { name: 'no-change', color: 'FF0000', description: '' },
+        { name: 'keep-me', color: '000000', description: '' },
+        { name: 'things-to-keep', color: '000000', description: '' },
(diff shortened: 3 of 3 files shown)


anderssonjohan:
ping @decyjphr @martinm82 Does this make sense? 

It occurred to me that the plugin is given the object from the config at the plugin level and not the whole config so instead of having a separate root config node it felt it made more sense to keep it in the labels plugin configuration rather than as a separate config key like I mentioned in the discussion thread.

It's only to exclude labels from being deleted but I thought using the `include` and `exclude` keys were the best way of turning the current array config to an object and making it similar to other parts of the configuration.

Other options considered and scrapped:

```
labels:
  exclude: regexp-pattern
  labels:
    - name: ....
      color: ...
```

```
labels:
  - exclude: regexp-pattern1
  - exclude: regexp-pattern2
  - name: label1
    color: ...
  - name: label2
    color: ...
```

Note: Unintentional white-space update of the README by my editor, let me know if you want me to redo the documentation update without the white-space update.

anderssonjohan:
Maybe an unexpected side-effect depending on the config values:
Using any of the following exclusion patterns will disable the deletion of any label, even the ones defined in `include`.

```yaml
labels:
  exclude:
    - .
    - '[^]'
```

martinm82:
> Maybe an unexpected side-effect depending on the config values: Using any of the following exclusion patterns will disable the deletion of any label, even the ones defined in `include`.
> 
> ```yaml
> labels:
>   exclude:
>     - .
>     - '[^]'
> ```

This is in my opinion a very nice addition and we should have this feature as well on topics (can be a separate PR if preferred). WDYT?

anderssonjohan:
I took the liberty to create a separate feature request for your suggestion about topics: https://github.com/github/safe-settings/issues/411

martinm82:
@anderssonjohan what do you think about documenting it in the README?

martinm82:
I was thinking a bit more about this and I am wondering whether the default behaviour should be to keep labels/topics and update with the ones managed in safe-settings.

In our case we have 9k repositories and started just recently using safe-settings. I cannot manage labels on org level since as of now it will delete all manually created ones. I believe that having the "do not delete" behaviour as default makes the transition to safe-settings much safer. WDYT?

Also, would be good to spit out in debug logs those labels that are not tracked by safe-settings. Later teams can get those from the logs (splunk, dataset, Loki, etc.) and migrate them over to safe-settings.

anderssonjohan:
> @anderssonjohan what do you think about documenting it in the README?

@martinm82 Do you mean documenting both syntax options? The (non-whitespace) change I made to the README in this PR was modifying the sample to use include/exclude for labels: https://github.com/github/safe-settings/pull/408/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L325

decyjphr:
@anderssonjohan this is a cool (and valuable) feature. Thank you. I'll test it and merge it.

martinm82:
> > @anderssonjohan what do you think about documenting it in the README?
> 
> @martinm82 Do you mean documenting both syntax options? The (non-whitespace) change I made to the README in this PR was modifying the sample to use include/exclude for labels: https://github.com/github/safe-settings/pull/408/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L325

Sorry, I missed that between all the whitespaces changes.

martinm82 (approved):
❤️

adrianschmidt:
> > > @anderssonjohan what do you think about documenting it in the README?
> > 
> > 
> > @martinm82 Do you mean documenting both syntax options? The (non-whitespace) change I made to the README in this PR was modifying the sample to use include/exclude for labels: [#408 (files)](https://github.com/github/safe-settings/pull/408/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L325)
> 
> Sorry, I missed that between all the whitespaces changes.

I think Johan is about to provide an updated version without the whitespace changes, so I'd wait for that before merging 🙂

anderssonjohan:
Sorry for the late response to this. Now ready for merge @decyjphr 

Thank you @adrianschmidt for pointing out that two trailing spaces means hard newline in Markdown!
I guess the (default) Markdown formatter used in VSCode has a [no hard-break-spaces rule](https://arcticicestudio.github.io/styleguide-markdown/rules/whitespace.html#trailing) and just strips them.
(The readme.md in this repo both have single and double spaces at EOLs)

Anyhow, readme now only have the intended change for the labels example.

EDIT: For the one also bitten by the VSCode formatter: https://github.com/microsoft/vscode/issues/1679#issuecomment-323608456

martinm82 (approved):

decyjphr:
@anderssonjohan for the existing configurations, Would this be a breaking change, i.e. they have to modify the `labels` section to move the current entries under the `includes:` subsection? Or would this change be backward compatible?

anderssonjohan:
> @anderssonjohan for the existing configurations, Would this be a breaking change, i.e. they have to modify the `labels` section to move the current entries under the `includes:` subsection? Or would this change be backward compatible?

@decyjphr Thanks for asking! it's backwards compatible by checking if the include is an array or not here: https://github.com/github/safe-settings/pull/408/files#diff-e71ad8761eefa8f71d7f5c23541f2078574ae83d48423be0d06dd16740d43cc3R8
Thus, it should be safe without being a breaking change.

martinm82:
> I was thinking a bit more about this and I am wondering whether the default behaviour should be to keep labels/topics and update with the ones managed in safe-settings.
> 
> In our case we have 9k repositories and started just recently using safe-settings. I cannot manage labels on org level since as of now it will delete all manually created ones. I believe that having the "do not delete" behaviour as default makes the transition to safe-settings much safer. WDYT?
> 
> Also, would be good to spit out in debug logs those labels that are not tracked by safe-settings. Later teams can get those from the logs (splunk, dataset, Loki, etc.) and migrate them over to safe-settings.

Actually, you can easily get this default behaviour by adding the exclude in the org settings.

martinm82:
@anderssonjohan I have been playing around with the exclude pattern (https://github.com/github/safe-settings/pull/408#issuecomment-1426193941) and it seems that if a repo is already managed with safe-settings and you create a label manually the first time it gets still deleted. The next time you create it again, safe-settings does not touch it again.
Have you seen this behaviour as well?

anderssonjohan:
> @anderssonjohan I have been playing around with the exclude pattern ([#408 (comment)](https://github.com/github/safe-settings/pull/408#issuecomment-1426193941)) and it seems that if a repo is already managed with safe-settings and you create a label manually the first time it gets still deleted. The next time you create it again, safe-settings does not touch it again. Have you seen this behaviour as well?

No, sorry, nothing that I've seen. Are you sure it got the webhook notification etc the second time?

PR #415, opened 2023-02-24
Fix up top-level array diffing

markjm:
Issue: When passing in 2 arrays to merge-deep, the diffing doesnt work as expected (due to array ordering). Converting arrays to an object-wrapped array works (i believe thats the intention of the old code that was there - but these objects still look like arrays)

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,10 @@ branches:
         dismissal_restrictions:
           users: []
           teams: []
+        bypass_pull_request_allowances:
+          users: []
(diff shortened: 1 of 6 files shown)


robandpdx:
@markjm Can you please describe what you mean by "the diffing doesn't work as expected". What are you expecting, and what are you seeing? These details will help me test this and get it merged. Thanks!

robandpdx (approved):

PR #454, opened 2023-04-06
Special handling in compareDeep for users, teams and apps

### LINKED ISSUE #433: Updating bypass_pull_request_allowances does not trigger webhook
## Problem Description
When you use the github UI and update the "Allow specified actors to bypass required pull requests" the setting will remain until the next time some other change is caught by the safe-settings event listeners.

When you change the bypass setting it is not overridden by the org setting.

When you change the bypass setting it should be overridden by the org setting.

I guess we're running our own?

GitHub App installed in GitHub Enterprise Cloud organization.

How do I know?

GitHub Enterprise Cloud

robandpdx:
This change fixes #433. The logic in mergeDeep.compareDeep() is a pain to understand and work with. It does a lot of work just to figure out what the additions and modifications are, only to report these details to the logs. Is that all necessary? I wan't going to rewrite this function for this change. 

The logic I implemented basically looks at both the source and target to find out of they differ from one another. If so, there is a change and the settings need to be reapplied. 

I have tested this and it seems to be working as expected.

### DIFF
--- a/lib/mergeDeep.js
+++ b/lib/mergeDeep.js
@@ -125,6 +125,8 @@ class MergeDeep {
                 visited[a.name] = a
               } else if (a.username) {
                 visited[a.username] = a
+              } else if (a.login) {
+                visited[a.login] = a
               }
             } else {
               // If already seen this, it is not a missing field
@@ -139,8 +141,28 @@ class MergeDeep {
           for (const fields of Object.keys(visited)) {
             combined.push(visited[fields])
           }
-          // Elements that are in target are not additions
-          additions[key] = combined.filter(value => !target[key].includes(value))
+          if (["apps","teams","users"].includes(key)) {
+            let sourceAdditions = combined.filter(value => !target[key].includes(value))
+            let targetAdditions = combined.filter(value => {
+              for (const entity of source[key]) {
+                if (entity.name === value.login) {
+                  return false
+                }
+                if (entity.username === value.login) {
+                  return false
+                }
+                if (entity.login === value.login) {
+                  return false
+                }
+              }
+              return true
+            }
+            )
+            additions[key] = sourceAdditions.length > targetAdditions.length ? sourceAdditions : targetAdditions
+          } else {
+            // Elements that are in target are not additions
+            additions[key] = combined.filter(value => !target[key].includes(value))
+          }
         } else {
           // recursively compare the objects
           this.compareDeep(target[key], source[key], additions[key], modifications[key])
@@ -158,6 +180,8 @@ class MergeDeep {
             modifications.name = source.name
           } else if (source.username) {
             modifications.username = source.username
+          } else if (source.login) {
+            modifications.login = source.login
           }
         }
       }
--- a/test/unit/lib/mergeDeep.test.js
+++ b/test/unit/lib/mergeDeep.test.js
@@ -221,6 +221,7 @@ branches:
     // console.log(`source ${JSON.stringify(source, null, 2)}`)
     // console.log(`target ${JSON.stringify(target, null, 2)}`)
     // console.log(`diffs ${JSON.stringify(merged, null, 2)}`)
+    expect(merged.hasChanges).toBeTruthy()
     expect(merged.additions).toEqual(expected.additions)
     expect(merged.modifications.length).toEqual(expected.modifications.length)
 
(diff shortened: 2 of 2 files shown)


decyjphr (commented):
I will look at it today @robandpdx . In the meantime, can you provide details on the use case for this change? Can I ask a favor? Try uncommenting the tests in `mergeDeep.test.js` and validate if they are passing. For e.g. `CompareDeep extensive test` could one uncommented and tested.

decyjphr:
@robandpdx The reason for figuring out changes is to avoid getting hit by rate-limits and also to make the app perform better for large installations. So when the settings file is changed for one particular thing, we only want to apply the settings that have changed for a repo.

robandpdx:
> I will look at it today @robandpdx . In the meantime, can you provide details on the use case for this change? Can I ask a favor? Try uncommenting the tests in `mergeDeep.test.js` and validate if they are passing. For e.g. `CompareDeep extensive test` could one uncommented and tested.

@decyjphr Currently there is a defect in this app. When users, teams or apps are changed in the UI for `dismissal_restrictions` or `bypass_pull_request_allowances` the change does not trigger safe-settings to restore the settings configured in the admin repo. Only after another change is made in the UI to trigger safe-settings to apply the configuration are these settings restored to the values configured in the admin repo. 

This change fixes the bug described above. So when a change to `dismissal_restrictions` or `bypass_pull_request_allowances` is made in the UI by adding or removing users, teams, or apps, safe-settings will immediately restore these settings to the values configured in the admin repo, as one would expect.

robandpdx:
@decyjphr All tests are passing...

jest test/unit/lib/mergeDeep.test.js
 PASS  test/unit/lib/mergeDeep.test.js
  ✓ CompareDeep does not mutate source object
  ✓ CompareDeep produces correct result for arrays of named objects
  ✓ CompareDeep result has changes when source is empty and target is not (1 ms)
  ✓ CompareDeep result has no change when source and target match
  ✓ CompareDeep finds modifications on top-level arrays with different ordering
  ✓ CompareDeep does not report changes for matching empty targets (1 ms)
  ✓ CompareDeep does not report changes for matching empty targets
  MergeDeep Tests
    ✓ CompareDeep extensive test (5 ms)
    ✓ CompareDeep Test when target is from the api

Test Suites: 1 passed, 1 total
Tests:       9 passed, 9 total
Snapshots:   0 total
Time:        0.198 s, estimated 1 s
Ran all test suites matching /test\/unit\/lib\/mergeDeep.test.js/i.

decyjphr (approved):
LGTM, I have a more general-purpose solution for this but we can tackle that in a different PR. I like what you did to unblock the users.

robandpdx:
> 

@decyjphr 
The mergeDeep.compareDeep() function returns additions, modifications, hasChanges. Each plugin only uses hasChanges to determine if changes are needed, then compares the settings from the admin repo with the repo settings to see what needs to be updated. For example...
https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/branches.js#L54
https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/repository.js#L91-L93

The additions and modifications objects are only used for log messages in these 2 places...
https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/branches.js#L64
https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/diffable.js#L85

It appears that at some point in the past additions and modifications object were used to determine if there were changes...
https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/repository.js#L73
... but this is now built into mergeDeep.compareDeep() as it returns the hasChanges boolean.

That said, the logic in mergeDeep.compareDeep() could be much simpler, as we really only care if something hasChanges or not.

decyjphr:
@robandpdx Then 2 main purposes for compareDeep  1. As you pointed out, to determine if there are indeed changes and only make the API calls if there are changes for a particular configuration, and 2. To determine the list of changes so that they are indicated in the PR workflow so the approver can understand the impact of changes before a change is applied to the main branch.

While there was a concrete need for it, I don't know who is using it. The PR flow needs more attention and the report that we have is not very helpful. The initial focus was to get the data and to work on the report later. 

The 3rd purpose for this is something that has come up recently: It might be useful if this info is persisted somewhere so that when some change was applied by accident, it is possible to rollback. It is not always possible to get this info from audit logs and events.

robandpdx:
@decyjphr 

> To determine the list of changes so that they are indicated in the PR workflow so the approver can understand the impact of changes before a change is applied to the main branch.

The changes in the PR itself should serve this purpose. No need to parse the data and report it. One can simply look at the changes in the PR to see the changes.

> The 3rd purpose for this is something that has come up recently: It might be useful if this info is persisted somewhere so that when some change was applied by accident, it is possible to rollback. It is not always possible to get this info from audit logs and events.

This information can be found in the git history and PR history of the repo. No need to report this anywhere else. To know what changed and when, one needs to inspect the repo history like any other codebase.

glyph-se:
> @decyjphr
> 
> > To determine the list of changes so that they are indicated in the PR workflow so the approver can understand the impact of changes before a change is applied to the main branch.
> 
> The changes in the PR itself should serve this purpose. No need to parse the data and report it. One can simply look at the changes in the PR to see the changes.

That is not always so easy to do. Especially if you are not aware exactly how safe-settings works.

Changing the settings of hundred of repos at the same time are something that can cause things to break for our users. I like how the safe-settings app lists what will be done so that you can know what the impact of merging a PR will be before it is applied.

This has saved us a few times with regards to how the structure of our teams works. (I'm talking in a general sense not the exact setting mentioned above)

PR #455, opened 2023-04-12
Use GitHub package registry. Build multiplatform image.

robandpdx:
This change will create multi-platform images and push them to GitHub container registry. This is preferred over @decyjphr's personal space in dockerhub.

I have tested the pre-release workflow [here](https://github.com/robandpdx/safe-settings-robandpdx/actions/runs/4673230985).


decyjphr (approved):
Good idea. 👍🏾

Issue #426, opened 2023-03-02
Please add Comments onto PR's / Commits with details about the operation.


justinmchase:
## New Feature

We are having an issue where settings are not being applied but are expected to be applied. There appears to be no error information available anywhere. Its not totally clear why its not working.

I know you won't always be able to provide information, such as if a suborg glob doesn't actually match then how would safe-settings be able to know that is an error? But in other cases, such as invalid settings file, it may be useful to provide more error information. Surfacing this information in an Action is ok but it would be nicer if information was attached directly to Commits / PR's in the form of Comments.

Therefore, please add:
1. A comment onto Commits / PR's which have error information whenever there is an invalid settings files in them.
2. A comment onto Commits in the default branch which lists all of the repositories that had their settings changed as a result of of the commit.

```[tasklist]

```

martinm82:
I had a similar issue and for that purpose wanted a pre-validation of the actual configuration file during a pull request, as these are usually the changes that result in issues with the GitHub REST API. 

Of course there are other issues if for example someone does not define `required_status_checks` in the branch protection rules.

What I did so far is implementing a schema validator to ensure the configuration files are properly defined. The validator as well ensures that org/suborg/repo files are properly merged.

I know that this is actually already part of safe-settings and could have been extracted into a separate module which we could call in a GitHub Action instead. I am not super fluent with JavaScript/TypeScript and implemented it therefore in Python.

Let me know if you are interested in this tool and I could create a branch on this repo here.

PR #479, opened 2023-05-14
Enable back the GitHub pull request comment with a feature flag

v1v:
### What

Add `ENABLE_PR_COMMENT` env variable to create GitHub comments on Pull Request contributions. There is already a feature to support a summary using the GitHub check commits. 

In this particular case I re-enable the existing commented code but using the feature flag `ENABLE_PR_COMMENT`

### Test

See https://github.com/elastic-robots/admin/pull/1#issuecomment-1546908435

### Issues

Notifies https://github.com/github/safe-settings/issues/426

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -122,10 +122,10 @@ class Settings {
     const template = eta.loadFile(path.join(__dirname, 'summary.tmpl'))
     const commentmessage = await eta.render(template, stats)
 
-    /*
-    const commentmessage2 = `
-## :robot: Safe-Settings Validator summary
-### Run on: ${new Date()}
+    if (env.ENABLE_PR_COMMENT === 'true') {
+      const commentmessage2 = `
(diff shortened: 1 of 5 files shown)


lib/settings.js
```diff
@@ -122,10 +122,10 @@ class Settings {
     const template = eta.loadFile(path.join(__dirname, 'summary.tmpl'))
     const commentmessage = await eta.render(template, stats)
 
-    /*
-    const commentmessage2 = `
-## :robot: Safe-Settings Validator summary
-### Run on: ${new Date()}
```
v1v, line 128: I removed this since the existing GitHub comment already contains when it was created

martinm82:
I guess we will hit quickly a rate limit if an org level change is discovered on an organization with 10k repositories.

decyjphr (approved):
Nice idea @v1v. I like what you have done with minimum code change.

PR #485, opened 2023-05-25
cosmetic: use table to report upcoming changes on PRs

v1v:
### What

Transformed the summary of changes in a HTML table.
I additionally changed the date and number of repos in a list of strings rather than headers, and added `None` if no errors

### Why

Help to visualise what repositories are affected for the new changes.

### UI changes

Current UI:

image

In this proposal

image

### Further examples

See https://github.com/elastic-robots/admin/pull/1/checks

### Questions

- If you like this approach please feel free to propose a different emojis, I looked at https://gist.github.com/rxaviers/7360908 but could not find something more explicit

### DIFF
--- a/lib/summary.tmpl
+++ b/lib/summary.tmpl
@@ -1,24 +1,41 @@
 ## :robot: Safe-Settings Settings Dry-Run Summary
-### Run on: <%= new Date() %>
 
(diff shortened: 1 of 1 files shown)


decyjphr (approved):
❤️ this change

Issue #502, opened 2023-07-18
Autolinks is_alphanumeric option is ignored


PiotrMarcinczyk-TomTom:
## Problem Description
We have an admin repository in the organization using safe-settings and wanted to apply [Autolinks](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources#about-autolinks) with **numeric** type because the alphanumeric ones are too eager in some cases.
For this, we prepared a suborg configuration with `autolinks` section as below:
```yaml
autolinks:
  - key_prefix: ABC-
    url_template: https://jira.tomtom.com/browse/ABC-<num>
    is_alphanumeric: false
```
The `is_alphanumeric` parameter was taken from [REST API reference](https://docs.github.com/en/rest/repos/autolinks?apiVersion=2022-11-28#create-an-autolink-reference-for-a-repository).

### What is actually happening
The `is_alphanumeric: false` parameter is ignored. Created autolinks are still alphanumeric.

### What is the expected behavior
The `is_alphanumeric: false` parameter is obeyed and created autolinks are of **numeric** type.

### Error output, if available
The related code seems to be here <https://github.com/github/safe-settings/blob/main-enterprise/lib/plugins/autolinks.js#L29> and looks like the `is_alphanumeric` setting is not handled there.

### Are you using the hosted instance of probot/settings or running your own?
hosted

### If running your own instance, are you using it with github.com or GitHub Enterprise?
github.com

#### Version of probot/settings
latest

```[tasklist]
### Tasks
```

PR #517, opened 2023-08-10
Support is_alphanumeric property in autolink configs

markjm:
Now that this is_alphanumeric property is in the API, we need to support it - especially because the default value (true) is a change from before its introduction

### DIFF
--- a/lib/plugins/autolinks.js
+++ b/lib/plugins/autolinks.js
@@ -13,11 +13,20 @@ module.exports = class Autolinks extends Diffable {
   }
 
   comparator (existing, attr) {
-    return existing.key_prefix === attr.key_prefix && existing.url_template === attr.url_template
+    return existing.key_prefix === attr.key_prefix &&
+      existing.url_template === attr.url_template
   }
 
   changed (existing, attr) {
-    return existing.key_prefix === attr.key_prefix && existing.url_template !== attr.url_template
+    // is_alphanumeric was added mid-2023. In order to continue to support settings yamls which dont specify this
+    // attribute, consider an unset is_alphanumeric as `true` (since that is the default value in the API)
+    // https://docs.github.com/en/rest/repos/autolinks?apiVersion=2022-11-28#create-an-autolink-reference-for-a-repository
+    const alphanumericMatches = attr.is_alphanumeric === undefined
+      ? existing.is_alphanumeric
+      : attr.is_alphanumeric === existing.is_alphanumeric
+    return existing.key_prefix === attr.key_prefix &&
+      existing.url_template !== attr.url_template &&
+      !alphanumericMatches
   }
 
   async update (existing, attr) {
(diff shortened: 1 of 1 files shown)


efenglu:
I believe this would fix #502

PiotrMarcinczyk-TomTom (approved):

PiotrMarcinczyk-TomTom:
Could you also add a few tests for this logic?

lib/plugins/autolinks.js
```diff
@@ -13,23 +13,33 @@ module.exports = class Autolinks extends Diffable {
   }
 
   comparator (existing, attr) {
-    return existing.key_prefix === attr.key_prefix && existing.url_template === attr.url_template
+    return existing.key_prefix === attr.key_prefix &&
+      existing.url_template === attr.url_template
```
stevoland, line 17: I realise you haven't changed this but it seems that since `key_prefix` must be unique that this comparison is more correct
```suggestion
    return existing.key_prefix === attr.key_prefix
```

lib/plugins/autolinks.js
```diff
@@ -13,23 +13,33 @@ module.exports = class Autolinks extends Diffable {
...
-    return existing.key_prefix === attr.key_prefix && existing.url_template !== attr.url_template
+    // is_alphanumeric was added mid-2023. In order to continue to support settings yamls which dont specify this
+    // attribute, consider an unset is_alphanumeric as `true` (since that is the default value in the API)
+    // https://docs.github.com/en/rest/repos/autolinks?apiVersion=2022-11-28#create-an-autolink-reference-for-a-repository
+    const isAlphaNumericMatch = attr.is_alphanumeric === undefined
+      ? existing.is_alphanumeric // === true, the default
+      : attr.is_alphanumeric === existing.is_alphanumeric
+    return existing.key_prefix === attr.key_prefix &&
+      existing.url_template !== attr.url_template &&
+      !isAlphaNumericMatch
```
stevoland, line 29: I think this condition is incorrect, should be
```suggestion
    return existing.url_template !== attr.url_template || !isAlphaNumericMatch
```

The `key_prefix` check in not required since this was already matched by the `comparator`?

stevoland (commented):
Thanks for this. I've made a couple of suggestions

lib/plugins/autolinks.js
decyjphr: Hi @markjm @stevoland should this stay open? Or should we implement the change above and merge it?

stevoland: Hi @decyjphr, we've been using the above with the changes suggested (with unit tests) in our fork for 6 months with no issues. I'd be happy to raise a PR next week.

We have a few more additions/fixes that we'd like to push upstream but would be more keen to contribute if the unit test at least were checked on CI.

decyjphr (approved):
Looks good to me

stevoland:
@decyjphr I raised https://github.com/github/safe-settings/pull/589 to supersede this one. I believe this one has a bug in the comparator.

decyjphr:
Thanks for letting me know. I'll correct it by merging #589

PR #519, opened 2023-08-17
Changes for handling org level rulesets

decyjphr:
[Rulesets](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-organization-settings/managing-rulesets-for-repositories-in-your-organization) are now available in GitHub. These changes are to support configuring Rulesets using safe-settings.

## Approach
We will support org level rulesets using `-rulesets:` entries in `.github/settings.yml`  and repo level rulesets using `-rulesets:` entries in the `.github/repos/<repo>.yml` (deferring the handling of rulesets entries in `<suborg>.yml` until ruleset targeting using repo properties is available and we would want to do the same in suborg yml)


decyjphr:
Added 2 env variables to control messages during errors or during the PR checkrun.
Env variable | default | behavior
-- | -- | --
CREATE_PR_COMMENT | true | Will show the differences in the PR comments between the existing config and the new changes being applied in the PR body
~CREATE_ERROR_ISSUE~ | ~true~ | ~Will create an issue in the `admin` repo when there are errors when applying the settings. So the errors don't happen silently~

Instead of this approach of creating an issue for showing the errors when safe-settings runs, I've made changes so a `check` is created for the latest commit. If there are errors the `check` shows as `failure`.

Previously when there were errors when applying safe-settings, or if the `custom-validators` return false, it is not visible to the users, with this approach, the `checkrun` details will show the errors when they occur.

PR #576, opened 2023-12-20
try without globalconfig and deploymentconfig

bheemvennapureddy:
try without globalconfig and deploymentconfig

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,10 +12,26 @@ let deploymentConfig
 module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) => {
   async function syncAllSettings (nop, context, repo = context.repo(), ref) {
     try {
-      deploymentConfig = await loadYamlFileSystem()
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -12,10 +12,26 @@ let deploymentConfig
 module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) => {
   async function syncAllSettings (nop, context, repo = context.repo(), ref) {
     try {
-      deploymentConfig = await loadYamlFileSystem()
-      robot.log.debug(`deploymentConfig is ${JSON.stringify(deploymentConfig)}`)
+      let deploymentConfig;
+      try {
+        deploymentConfig = await loadYamlFileSystem();
+      } catch (error) {
+        // Assumption here is if some one is using just repo config they don't need include exclude list of repos
```
svg153, line 19: Wouldn't it be better to create a feature flag, instead of assuming this functionality? Assuming it can create confusion and not leave the execution logic clear. What is your opinion ?

bheemvennapureddy: ya agree

PR #580, opened 2024-01-03
Create official helm chart for safe-settings

### LINKED ISSUE #578: Create official helm chart
## New Feature

Currently, there is a [helm chart](https://github.com/decyjphr-org/charts/tree/main/helm-charts/safe-settings) linked in the deployment docs.
But it feels a bit odd to use a helm chart that isn't managed by the github org.
Also, the helm chart haven't gotten any release since 2022 https://decyjphr-org.github.io/charts/index.yaml.

There were an initial PR for this back in 2022, but it was [closed](https://github.com/github/safe-settings/pull/261) for some reason.

Since helm now supports OCI we don't have to setup any strange github page to be able to pull the chart.
If this is something that you would be up for, I'm happy to create a PR for it + a GitHub action that publishes the chart on a release.

nissessenap:
This helm chart has taken much inspiration from https://github.com/decyjphr-org/charts/tree/main/helm-charts/safe-settings, but it differs in one main way.  It don't force the user to put any secrets directly in the values files. Instead, I make it possible to either set the environment variables or point to any Kubernetes secret you want. I assume that the secret already exists, how this secret is up to the user of the helm chart.

In a GitOps world, I would suggest something like 

- sealed secrets
- external-secrets 
- etc.

Or if you want, you can still add the secret through Kubernetes.

This of course makes it harder to use the helm chart, but it will help to enforce best practices. We can of course extend this PR to support a similar Kubernetes secret injection that the current helm chart supports. But it needs to be able to disable this feature.
 
I have created a helm chart so you can try it out if you want `helm upgrade -i safe-settings oci://ghcr.io/nissessenap/helm-charts/safe-settings --version 2.1.0`

## TODO

- I haven't written any docs
- The release won't work due to how the current release workflow works. This is an example on how it could be done.
If we instead would trigger on release https://github.com/NissesSenap/safe-settings/blob/main-enterprise/.github/workflows/create-release-2.yml this would work, but I don't want to put too much time in to this unless someone thinks this is worth merging.

Partly fixes #578 

Looking forward to your feedback.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -66,6 +66,27 @@ jobs:
           provenance: false
           sbom: false
 
+  helm:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v4
+
+      - name: Prepare
+        run: |
+          # OCI standard enforces lower-case paths
+          GHCR_REPO=$(echo "ghcr.io/${{ github.repository }}" | tr '[:upper:]' '[:lower:]')
+          HELM_REPO=$(echo "oci://ghcr.io/${{ github.repository_owner }}/helm-charts" | tr '[:upper:]' '[:lower:]')
+          echo "GHCR_REPO=$GHCR_REPO" >> $GITHUB_ENV
+          echo "HELM_REPO=$HELM_REPO" >> $GITHUB_ENV
+
+      - name: Publish Helm charts
+        run: |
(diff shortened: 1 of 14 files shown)


nissessenap:
Should be able to use `/probot` and health endpoint

decyjphr:
Thanks @NissesSenap I'll do a test deploy and then merge. Awesome that we will have a helm chart that is kinda official.

PR #582, opened 2024-01-12
Fix Lambda/Serverless Cron

benbridts:
This fixes a bug introduced by changing module.exports in index.js
(commit 1540810)

In https://github.com/github/safe-settings/commit/15408100d15ac18d2698ac2e3bd901b6ba5545db the following change happend in index.js (and that broke the cron/scheduler handler):

```diff
- module.exports = (robot, _, Settings = require('./lib/settings')) => {
+ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) => {
```

### DIFF
--- a/handler.js
+++ b/handler.js
@@ -11,6 +11,6 @@ module.exports.webhooks = createLambdaFunction(appFn, {
 
 module.exports.scheduler = function () {
   const probot = createProbot()
-  const app = appFn(probot)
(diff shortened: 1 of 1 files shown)


benbridts:
@decyjphr : sorry for the tag, but I couldn't find the best way to get this PR discussed. Is this a change you would consider merging?

benbridts:
❤️

PR #584, opened 2024-02-01
Allow github context in configValidator and overrideValidator

ghost:
- Added new property (github) in MergeDeep Constructor
- Added github context as argument whenever isValid and canOverride functions are invoked
- Added 'githubContext' as a new parameter whenever a new function is created

We also tested our changes to see if githubContext was available to use via deployment-settings.yaml :

``` 
restrictedRepos:
  # You can exclude certain repos from safe-settings processing
  #  If no file is specified, then the following repositories - 'admin', '.github', 'safe-settings' are exempted by default
  exclude: ['^admin$', '^\.github$', '^safe-settings$', '.*-test']
  # # Alternatively you can only include certain repos
  # include: ['^test$']
configvalidators:
  - plugin: repository
    error: |
      `Repository plugin errored`
    script: |
      githubContext.request('/repos/{owner}/{repo}', { owner: 'cdot-dev' , repo: 'dauxs-cdot-demo' }).then(res => { console.log(res.data) })

      if (new RegExp(baseconfig.validator.pattern).test(baseconfig.name)){
        console.log(`${baseconfig.name} matches ${baseconfig.validator.pattern}`)
      }
      return true
```

and we were able to successfully get a response back.

Edit: Fixed mergeDeep.test.js (mock function returns an anonymous function)

### DIFF
--- a/app.yml
+++ b/app.yml
@@ -135,4 +135,4 @@ url: https://github.com/github/safe-settings
 
 # Set to true when your GitHub App is available to the public or false when it is only accessible to the owner of the app.
 # Default: true
-public: false
+public: true
--- a/app.yml
+++ b/app.yml
@@ -135,4 +135,4 @@ url: https://github.com/github/safe-settings
 
 # Set to true when your GitHub App is available to the public or false when it is only accessible to the owner of the app.
 # Default: true
-public: true
+public: false
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "eslint-plugin-node": "^11.1.0",
         "eslint-plugin-promise": "^6.0.1",
         "http-status-codes": "^2.2.0",
-        "jest": "^29.5.0",
+        "jest": "^29.7.0",
         "jest-when": "^3.5.2",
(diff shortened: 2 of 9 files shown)


package-lock.json
decyjphr, line 1: Hi, Could you delete the package-lock.json locally and run npm install again to remove any refs to fdic and packages like word-wrap which are not in package.json

decyjphr (changes requested):

package-lock.json
decyjphr: Thanks. Got the latest code and ran tests locally. 👍🏾

decyjphr (approved):

app.yml
```diff
@@ -135,4 +135,4 @@ url: https://github.com/github/safe-settings
 
 # Set to true when your GitHub App is available to the public or false when it is only accessible to the owner of the app.
 # Default: true
-public: false
```
decyjphr: Could we keep this as private? i.e. keep the default posture as app being private

decyjphr (changes requested):

decyjphr (approved):

PR #351, opened 2022-12-05
feat: allow different admin repo by env variable

quike:
Issue: [230](https://github.com/github/safe-settings/issues/230)

Extract configuration to an external library so it can be updated via env variables. This should allow to select a different default admin organization.

This change permits to pass an environment variable for `ADMIN_REPO`, `SETTINGS_FILE_PATH` and `DEPLOYMENT_CONFIG_FILE` so they can be modified, like for example selecting as `ADMIN_REPO` the repository `.github` as other probot apps.

| Environment Variable               | Internal Reference       | Default Value             |
|------------------------------------|--------------------------|---------------------------|
| `ADMIN_REPO`         | `ADMIN_REPO`             | `admin`                   |
| `SETTINGS_FILE_PATH` | `SETTINGS_FILE_PATH`     | `settings.yml`            |
| `DEPLOYMENT_CONFIG_FILE`           | `DEPLOYMENT_CONFIG_FILE` | `deployment-settings.yml` |

Example:

```
ADMIN_REPO=global-config
SETTINGS_FILE_PATH=safe-settings.yml
DEPLOYMENT_CONFIG_FILE=safe-settings-deployment.yml
```

And the code will look within your `<ORG>/global-config/.github/safe-settings.yml` instead of default `<ORG>/admin/.github/settings.yml`

PR #588, opened 2024-02-14
Adding optional CONFIG_PATH env var support

dolan-a:
This PR adds support for defining the CONFIG_PATH settings through an environment variable. I'm not in love with the env var's name since it is pretty generic, but I stuck with the naming convention of the other variables in env.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@
    1. Org-level settings are defined in `.github/settings.yml`  
        > **Note**
        > It is possible to override this behavior and specify a different filename for the `settings` yml repo.<br>
-       > This could be done by setting an `env` variable called `SETTINGS_FILE_PATH`.
+       > This could be done by setting an `env` variable called `SETTINGS_FILE_PATH`.<br>
+       > Similarly, the `.github` directory can be overridden with an `env` variable called `CONFIG_PATH`.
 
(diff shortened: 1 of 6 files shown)


decyjphr:
Strange - this PR is basically undoing a previous change https://github.com/github/safe-settings/commit/18509385eaae9762379a2884e4491e60d5ae6bf3 which removed this flag.

dolan-a:
Thanks @decyjphr for pointing out the other PR. I looked at the comments on #351, and I think my PR is still worth considering. A few specific points to address:

- **Probot's suggestion of using `.github` for config files:** As @martinm82 noted, Probot's best-practices docs say to store configs in the `.github` directory of repos. I think this suggestion is more appropriate for GitHub Apps that would load repo-specific configurations from the repo directly. Since `safe-settings` instead stores all configurations in an "admin" repo, I don't think this best-practice is as relevant. 

- **Updating all hardcoded ".github" directory references:** The author of #351 rightly noted that there are several spots in the code that are hardcoding the config-path directory. I've since updated this PR to use `env.CONFIG_PATH` instead. I left most of the README as is -- other than adding a note about the CONFIG_PATH env being an optional param. 


Ideally we'd have some unit tests to ensure that the repo and sub-org overrides are being loaded correctly when overriding CONFIG_PATH. It wasn't clear to me how best to go about this though, so if anyone has suggestions, let me know.

decyjphr (approved):
Thanks for the changes.

PR #594, opened 2024-02-15
fix(branches): Users in branch protections always deleted and added

stevoland:
`type` property ("User") is matched as the identifier before `login` so the diff always adds addition/deletion.

Theoretically this fix could break some other diff use-case and I wouldn't know


decyjphr (approved):
👍🏾

PR #596, opened 2024-02-17
fix(branches): Not adding checks to existing branch protection contexts

stevoland:
All I can say is that the changed line implements what I assume is the intention of mutating the `modifications` parameter and fixes this particular case.


stevoland:
My bad,  my unit test was incorrect and the problem remains

Issue #379, opened 2023-01-12
Trigger manually sync/full sync


martinm82:
## Prerequisites:

* Is the functionality available in the GitHub UI? If so, please provide a link to information about the feature.

* Is the functionality available through the GitHub API? If the functionality is available, please provide links to the
  API documentation (https://developer.github.com/v3/) as well as the Octokit documentation (https://octokit.github.io/).

* If the functionality is not yet available in the API, it would be helpful if you
  contacted support (https://support.github.com/) or posted in the Community Forum (https://github.community/). Please
  include a link to the forum post if you create one or a copy of the response from support.

## New Feature

Imagine you have the situation where the safe-settings app did not configure a repository properly, because for whatever reason, e.g., webhook not send to the app, an application error, etc.. 

Wouldn't it be great if there would be at least the possibility to trigger manually a sync or full sync through a dedicated endpoint?

decyjphr:
This was solved for lambda by exporting `syncInstallation` method in the index.js https://github.com/github/safe-settings/issues/313

Alternatively, we can create an HTTP endpoint in the app that would do the same.

anderssonjohan:
@martinm82 I'm thinking I will try to set up a GitHub Actions workflow using cron + workflow dispatch (manual run) to call the lambda function.
I have just started using this app but find it hard to get feedback on what happened for the real run after PRs are merged to the main branch of the admin repo. It would be great to be able to re-run the changes on `main` and get feedback from the lambda back as a commit status. The dry run is enough for PRs but is not good enough for `main` IMHO.

dolan-a:
As noted on #378, I have a prototype runner that I'm using for running safe-settings via GHA. If we like this approach of calling `syncInstallation` via a separate script, ~I can open a PR (was just delaying since I forked off my changes in #588)~ *update: PR opened in #604*

Issue #378, opened 2023-01-11
Possibility to run safe-settings as GitHub Action?


martinm82:
## Prerequisites:

* Is the functionality available in the GitHub UI? If so, please provide a link to information about the feature.

* Is the functionality available through the GitHub API? If the functionality is available, please provide links to the
  API documentation (https://developer.github.com/v3/) as well as the Octokit documentation (https://octokit.github.io/).

* If the functionality is not yet available in the API, it would be helpful if you
  contacted support (https://support.github.com/) or posted in the Community Forum (https://github.community/). Please
  include a link to the forum post if you create one or a copy of the response from support.

## New Feature

Based on the Probot docs there is an GitHub Action adapter which should allow running Probot apps as Actions. Would this be possible as well with safe-settings?

https://probot.github.io/docs/deployment/#github-actions

niraj8:
There are 2 ways I think we could run this in Github Actions
- On a schedule that does a full reconciliation , i.e. check that all settings across all repositories in the org match the desired settings from the `admin` repo, similar to #313 
- Trigger a GH workflow that runs `safe-settings` app with the webhook payload event to a Github workflow with `inputs` in `workflow_dispatch` [[example]](https://www.jfrog.com/confluence/display/JFROG/Trigger+a+GitHub+Action+using+Custom+Webhook)

dolan-a:
I have a prototype of this on my repo fork, which adds a new script for calling `syncInstallation` (if people like this approach, ~I can open a PR, but I'll have to separate it from my changes on #588~ *update: PR opened in #604*):

```diff
diff --git a/full-sync.js b/full-sync.js
new file mode 100644
index 0000000..7881056
--- /dev/null
+++ b/full-sync.js
@@ -0,0 +1,6 @@
+const { createProbot } = require('probot')
+const appFn = require('./')
+
+const probot = createProbot()
+const app = appFn(probot, {})
+app.syncInstallation()
diff --git a/package.json b/package.json
index 6bfb4ce..624f429 100644
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "dev": "nodemon --inspect",
     "start": "probot run ./index.js",
+    "full-sync": "node ./full-sync.js",
     "test": "npm-run-all --print-label --parallel lint:* --parallel test:*",
     "lint:es": "eslint .",
     "lint:js": "standard",
``` 

[...]

PR #604, opened 2024-02-28
Adding external runner for doing a one-time full-sync

dolan-a:
### Summary: ###
This is a potential way of addressing #378 and #379. 

As noted on #378, I'm using this along with the following for triggering safe-settings via GHA:
```yaml
name: Safe Settings Sync
on:
  schedule:
    # daily run:
    - cron:  '0 0 * * *'
  workflow_dispatch: {}

jobs:
  safeSettingsSync:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          repository: pydolan/safe-settings
          ref: gha-runner
      - uses: actions/setup-node@v4
      - run: npm install
      - run: npm run full-sync
        env:
          GH_ORG: my-org
          APP_ID: my-app-id
          PRIVATE_KEY: ${{ secrets.SAFE_SETTINGS_PRIVATE_KEY }}
          GITHUB_CLIENT_ID: ${{ secrets.SAFE_SETTINGS_GITHUB_CLIENT_ID }}
          GITHUB_CLIENT_SECRET: ${{ secrets.SAFE_SETTINGS_GITHUB_CLIENT_SECRET }}
```

### Additional comments: ###
- My Commit history is a mess, so if you weren't going to do a Squash Merge of this PR, I will clean this up.
- If there's a more javascript-y way of calling `syncInstallation` using index.js/package.json, I'm happy to change this.
- Regarding [Probot's GHA Adapter](https://github.com/probot/adapter-github-actions) -- I initially used this in my separate script (similar to what handler.js does with the Serverless adapter), but this adapter uses the Action's GITHUB_TOKEN, which is limited to the current repo, so it offers no benefit that I can see.
- Regarding my use of actions/checkout -- I would prefer to run safe-settings as an action using the Dockerfile, but GHA targets a different WORKDIR when doing so. There's an https://github.com/actions/runner/issues/878 about allowing workdir overrides with GHA.
- ***[As noted in comments]*** I realized that my initial version of full-sync did not return a non-zero exit code when errors occurred, so I updated it to check `settings.errors`. I didn't see a way to check the errors from my syncInstallation call, so I modified `Settings.syncAll` to return its `settings` object. If someone has a cleaner option for this error checking, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,11 @@
 >
 > Settings files must have a `.yml` extension only. For now, the `.yaml` extension is ignored.
 
+
 ## How it works
+
+`Safe-settings` is designed to run as a service listening for webhook events or as a scheduled job running on some regular cadence. It can also be triggered through GitHub Actions. (See the [How to use](#how-to-use) section for details on deploying and configuring.)
+
 
 ### Events
 The App listens to the following webhook events:
@@ -364,11 +368,13 @@ You can pass environment variables; the easiest way to do it is via a `.env` fil
 
 ## How to use
 
-1. __[Deploy and install the app](docs/deploy.md)__.
+1. Create an `admin` repo (or an alternative of your choosing) within your organization. Remember to set `CONFIG_REPO` if you choose something other than `admin`. See [Environment variables](#environment-variables) for more details.
 
-2. Create an `admin` repo (or an alternative of your choosing) within your organization. Remember to set `CONFIG_REPO` if you choose something other than `admin`. See [Environment variables](#environment-variables) for more details.
+2. Add the settings for the `org`, `suborgs`, and `repos`. Sample files can be found [here](docs/sample-settings).
 
-3. Add the settings for the `org`, `suborgs`, and `repos`. Sample files can be found [here](docs/sample-settings).
+3. __[Deploy and install the app](docs/deploy.md)__.  Alternatively, the __[GitHub Actions Guide](docs/github-action.md)__ describes how to run `safe-settings` with GitHub Actions.
+
+
 
 
 ## License
--- a/docs/github-action.md
+++ b/docs/github-action.md
@@ -1,0 +1,57 @@
+# Running Safe-settings with GitHub Actions (GHA)
+
+This guide describes how to schedule a full safe-settings sync using GitHub Actions. This assumes that an `admin` repository has been configured with your `safe-settings` configuration. Refer to the [How to Use](../README.md#how-to-use) docs for more details on that process.
+
+
+## GitHub App Creation
+Follow the [Create the GitHub App](deploy.md#create-the-github-app) guide to create an App in your GitHub account. This will allow `safe-settings` to access and modify your repos.
+
+
+## Defining the GitHub Action Workflow
+Running a full-sync with `safe-settings` can be done via `npm run full-sync`. This requires installing Node, such as with [actions/setup-node](https://github.com/actions/setup-node) (see example below). When doing so, the appropriate environment variables must be set (see the [Environment variables](#environment-variables) document for more details).
+
+
+### Example GHA Workflow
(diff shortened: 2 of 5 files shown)


dolan-a:
I realized that my initial version of full-sync did not return a non-zero exit code when errors occurred, so I updated it to check `settings.errors`. I didn't see a way to check the errors from my syncInstallation call, so I modified `Settings.syncAll` to return its `settings` object. 

If someone has a cleaner option for this error checking, please let me know

decyjphr:
We could look into using ___checks___ to communicate errors back to the Actions workflow.

dolan-a:
Yea, tying into Checks would make things cleaner. I'm not familiar with the feature or code in detail, so if anyone has tips/suggestions, let me know.

The other question I had was if the probot gha adapter would help with error collecting, but it doesn't look like the Settings class is sending probot error info, so maybe it wouldn't help

decyjphr (commented):
@pydolan I think this is a cool addition. Would you be able to add information about this in the README?

dolan-a:
@decyjphr - this is still on my radar; will try to get to it soon!

willejs:
@decyjphr @decyjphr it would be great to see this merged in! ❤️

decyjphr (approved):

dolan-antenucci-imprivata:
@decyjphr and others – thanks for your patience on me getting back to this. I rebased my branch and updated the documentation with details on running `safe-settings` with GHA.

paddyroddy:
@dolan-a thanks for sorting this! I'm deploying `safe-settings` via the GHA. I had a play with [sample deployment settings](https://github.com/github/safe-settings/blob/main-enterprise/docs/sample-settings/sample-deployment-settings.yml) and wasn't sure if the `configvalidators`/`overridevalidators` sections were actually working. I noticed this bit in the `README.md`. To enable the runtime settings, must one deploy, e.g. the `helm` chart?

> ### Runtime Settings
> 
> 3. Currently the only setting that is possible are `restrictedRepos: [... ]` which allows you to configure a list of repos within your `org` that are excluded from the settings. If the `deployment-settings.yml` is not present, the following repos are added by default to the `restricted`repos list: `'admin', '.github', 'safe-settings'`

dolan-antenucci-imprivata:
I'm using a custom runtime settings file in my GHA setup to exclude repos. I would guess that the validators are working from my file (I left the default ones in there from the example file), but I forget if I tested them.

To use this, as I show in  [https://github.com/github/safe-settings/blob/main-enterprise/docs/github-action.md](this example GHA setup), I have a var for indicating where that setting file is:
```yaml
DEPLOYMENT_CONFIG_FILE: ${{ github.workspace }}/safe-settings/deployment-settings.yml
```

This does not require the helm deployment since the entire safe-settings code is pulled down to the GHa runner, and they code is what uses the deployment settings env var. Hope that helps!

dolan-a:
I'm using a custom runtime settings file in my GHA setup to exclude repos. I would guess that the validators are working from my file (I left the default ones in there from the example file), but I forget if I tested them.

To use this, as I show in  [https://github.com/github/safe-settings/blob/main-enterprise/docs/github-action.md](this example GHA setup), I have a var for indicating where that setting file is:
```yaml
DEPLOYMENT_CONFIG_FILE: ${{ github.workspace }}/safe-settings/deployment-settings.yml
```

This does not require the helm deployment since the entire safe-settings code is pulled down to the GHa runner, and they code is what uses the deployment settings env var. Hope that helps!

paddyroddy:
Thanks @dolan-a! I also had the default ones, I tried changing a collaborator to `admin` and it didn't seem to do anything (including manually running the action), which is what made me wonder if it's working. I, too, am using the `restrictedRepos` which is working well.

decyjphr:
@paddyroddy I created this [issue](https://github.com/github/safe-settings/issues/715) to check if you use case still works.

PR #617, opened 2024-04-05
Add code to revert back repo renames

decyjphr:
When a repo is renamed outside safe-settings we want to revert it back to the original name.

We get the `repository.renamed` event and the payload has the previous name in it. 

Then, we can simply call `Settings.sync` with the repo settings for the old name. This will revert the changes  because when we call the api to get a repo with the `old` name it will retrieve the repo with the `new` name automatically. And the rest is easy as if we are applying the settings to the repo normally with repo name as the `old` name.


decyjphr:
Adding this functionality does add complications. 
1. First, if `safe-settings` prevents renaming of repos by humans, then it should allow for a way to rename repos using safe-settings itself. For this, the logical thing to do is to allow users to request a repo rename by renaming the `repo/<repo>.xml`. This is implementable because, when a push happens we can detect rename of `<repo>.xml` and do the rename operation. Similarly, in the PR workflow, we can detect a rename and report that as a change.
2. We don't want to block existing users of `safe-settings` by suddenly reverting repo renames in their orgs. So I am thinking of enabling this feature only when it is explicitly enabled in the `<deployment-config>` like using a setting ___block_repo_rename=true___

I am interested in your thoughts. Please let me know.

decyjphr:
Added code so that:

If a repository is renamed, the default behavior is safe-settings will ignore this (for backward-compatibility). 

If `BLOCK_REPO_RENAME_BY_HUMAN` env variable is set to true, `safe-settings` will revert the repo to the previous name unless it is renamed using a `bot`. 

If it is renamed using a `bot`, it will try to copy the existing `<old-repo>.yml` to `<new-repo>.yml` so that the repo config yml stays consistent. If the <new-repo.yml> file is already existing, it doesn't create a new one.

If you rename a <repo.yml> that corresponds to a repo, safe-settings will rename the repo to the new name. This behavior will take effect whether the env variable `BLOCK_REPO_RENAME_BY_HUMAN` is set or not.

PR #624, opened 2024-04-26
Ability to create suborgs using custom properties.

decyjphr:
Added support to create definitions of `suborgs` based on custom properties.

You can use custom properties to create a suborg in the `.github/suborgs/<suborg>.yml` as follows:
```yaml
suborgproperties:     
- prop1: true
- prop2: frontend

...
```

In order for this to work, the app has to be given `organization_custom_properties: admin` permissions and it has to listen to `custom_property_values` event.

### DIFF
--- a/docs/sample-settings/suborg.yml
+++ b/docs/sample-settings/suborg.yml
@@ -5,8 +5,13 @@ suborgrepos:
   - test*
 # You can use Glob patterns
 
+# List of repos that belong to the suborg
 suborgteams:
   - core
+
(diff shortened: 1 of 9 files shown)


docs/sample-settings/suborg.yml
```diff
@@ -5,9 +5,14 @@ suborgrepos:
   - test*
 # You can use Glob patterns
 
+# List of repos that belong to the suborg
```
jacobevansgit, line 8:
```suggestion
# List of teams that belong to the suborg
```

PR #626, opened 2024-05-09
feat: repository custom properties plugin

### LINKED ISSUE #543: Repository Custom Properties
## Prerequisites:

* Is the functionality available in the GitHub UI? If so, please provide a link to information about the feature.
https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization

* Is the functionality available through the GitHub API? If the functionality is available, please provide links to the
  API documentation (https://developer.github.com/v3/) as well as the Octokit documentation (https://octokit.github.io/).
https://docs.github.com/en/rest/repos/properties?apiVersion=2022-11-28
https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28

## New Feature

Support configuring the new repository custom properties in a repository config. This feature is really useful to setup different rulesets according to language, environments, compliance requirements etc. I already have a client that would use this feature and would prefer to be able to do it through safe-settings.

It could look something like this:
```yaml
repository:
  custom_properties:
    property1: value1
    property2: value2
    ...
```

If the property does not already exists at the org level, it should create it before setting the property. Only supporting text properties is fine since input validation can be done on the safe-settings files themselves.

Simon-Boyer:
# Repository Custom Properties

Fixes https://github.com/github/safe-settings/issues/543.

This new plugin allows to define custom properties values using this API: https://docs.github.com/en/rest/repos/custom-properties.
The syntax is:
```yaml
custom_properties:
- name: custom_prop1
  value: foo
- name: custom_prop2
  value: bar
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,12 @@ branches:
         users: []
         teams: []
 
+# Custom properties
+# See https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28
+custom_properties:
+  - name: test
+    value: test
+
 # See the docs (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources) for a description of autolinks and replacement values.
 autolinks:
   - key_prefix: 'JIRA-'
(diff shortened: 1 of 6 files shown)


Simon-Boyer:
@decyjphr do you think you'll be able to look at this PR soon? I would actually need this for a project I am currently working on. Thanks a lot!

PR #635, opened 2024-05-26
Feature: Add support for repository variables

primetheus:
This PR adds support for repository-level variables that can be used outside of a specific deployment environment. 

cc: @decyjphr


lib/plugins/variables.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /**
+     * Look-up existing variables for a given repository
+     *
+     * @see {@link https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28#list-repository-variables} list repository variables
+     * @returns {Array.<object>} Returns a list of variables that exist in a repository
+     */
+  find () {
+    const result = async () => {
```
decyjphr, line 25: The code was failing the tests, perhaps the find method could be rewritten as
```suggestion
  async find () {
    const res = await this.github.request('GET /repos/:org/:repo/actions/variables', {
      org: this.repo.owner,
      repo: this.repo.repo
    })

    const properties = res.data.map(d => { return { name: d.property_name, value: d.value } })
    return properties
  }
```

PendaGTP:
@primetheus could you also update the related docs? Like provide examples to make the distinction between environment variables and GH Actions variables more clear (#820)?

Issue #628, opened 2024-05-14
Environment reviewers must have at least one reviewer to set prevent_self_review


gregnrobinson:
## Problem Description
 
When creating environments that have reviewers set and the `prevent_self_review` setting to `true`, safe-settings does not process all environments correctly when first creating the environments and complains that the reviewer's section is not set even though it is configured correctly. On the second pass, it will create the rest of the `prod` environment which failed to provision correctly.
 
### What is actually happening
 
Safe settings will create the three environments with no reviewers and then error out on the `prod` environment leaving it partially provisioned. The pod running safe-settings will restart and when restarted it will finish provisioning the prod environment correctly.
 
### What is the expected behavior
 
Safe settings should create all the environments when triggered to do so without any errors or pod restarts.
 
### Error output, if available
 
```
Error HttpError: Failed to create or update the environment protection rule. Required reviewers must have at least one reviewer to set prevent_self_review. in Environments for repo: {"owner":"XXX","repo":"XXXXXXX","branch":"master"} entries [{"name":"dev","wait_timer":0,"prevent_self_review":false,"reviewers":[],"deployment_branch_policy":{"protected_branches":false,"custom_branch_policies":[]},"variables":[{"name":"base_environment","value":"DEV"}],"deployment_protection_rules":[]},{"name":"qat","wait_timer":0,"prevent_self_review":false,"reviewers":[],"deployment_branch_policy":{"protected_branches":false,"custom_branch_policies":[]},"variables":[{"name":"base_environment","value":"QAT"}],"deployment_protection_rules":[]},{"name":"uat","wait_timer":0,"prevent_self_review":false,"reviewers":[],"deployment_branch_policy":{"protected_branches":false,"custom_branch_policies":[]},"variables":[{"name":"base_environment","value":"UAT"}],"deployment_protection_rules":[]},{"name":"prod","wait_timer":0,"prevent_self_review":true,"reviewers":[{"type":"Team","id":17632}],"deployment_branch_policy":{"protected_branches":true,"custom_branch_policies":false},"variables":[{"name":"base_environment","value":"PROD"}],"deployment_protection_rules":[]}]
```
 
## Context
 
When we first started configuring environments we noticed there were a lot of errors related to attributes like `variables`, `deployment_protection_rules` and `custom_branch_policies` not being iterable or simply not having these attributes declared in the configuration causing various errors. So the final configuration we landed on was the one below which declares all the settings and if we do not want to apply those settings we leave the value with `[]`.

[...]

Issue #623, opened 2024-04-18
Environments do not get provisioned for repositories set to internal or private


gregnrobinson:
## Problem Description
 
When managing environments with safe settings, if the repository is set to public, safe settings will create all environments defined within the suborg configuration. If the repository visibility is set to **_private_** or **_internal_**, safe settings presents an error even though the provisioning of environments has been tested for all repository visibilities via the REST API.
 
### What is actually happening
 
If the repositories are anything but public visibility, safe settings cannot provision the environments.
 
### What is the expected behavior
 
Safe Settings should be able to manage environments on repositories that are set to private or internal visibility.
 
### Error output, if available
 
```
Error HttpError: Resource not accessible by integration in Environments for repo. {team details...}
```
 
## Context
 
We had initially opened #611 regarding environment provisioning but assumed the issue was due to insufficient licensing for the non-prod environments but after deploying to production, the environment provisioning would still only work for public repositories.
 
Using the REST API, we can successfully create environments for repositories even if the repository is set internal or private visibility. We are wondering why Safe settings specifically returns an error for environments that are being provisioned against internal or private repositories.
 
### Create environment for a public repository
 
```bash
curl -L \
-X PUT \
-H "Accept: application/vnd.github+json" \
-H "Authorization: Bearer $token" \
-H "X-GitHub-Api-Version: 2022-11-28" \
[https://%3cGHES_ENDPOINT%3e/api/v3/repos/%3cORG%3e/%3crepoprefix%3e-iam-refactoring/environments/dantestpublic]https://<GHES_ENDPOINT>/api/v3/repos/<ORG>/<repoprefix>-iam-refactoring/environments/dantestpublic \
-d '{"wait_timer": 0,"prevent_self_review": false,"reviewers": [],"deployment_branch_policy": null}'
```
 
**response**
```json
{
"id": 3523,
"node_id": "MDExOkVudmlyb25tZW50MzUyMw==",
"name": "dantestpublic",
"url": [https://%3cGHES_ENDPOINT%3e/api/v3/repos/%3cORG%3e/%3crepoprefix%3e-iam-refactoring/environments/dantestpublic]https://<GHES_ENDPOINT>/api/v3/repos/<ORG>/<repoprefix>-iam-refactoring/environments/dantestpublic,
[...]

PR #616, opened 2024-04-04
plugins/environments.js: Fix creation of new environments

klutchell:
Currently if environments do not exist in the target repository, they cannot be created
if all properties are not populated.

This PR performs a lint of the environments plugin file (first commit) before adding new
test cases to cover the creation of environments that do not exist.

The command `eslint --fix ./lib/plugins/environments.js`
was used to lint the plugin and some variable names were
updated to camel case manually.

PR #646, opened 2024-06-18
Improve change detection for environments, expand unit tests

Brad-Abrams:
## Environments unit test hid an error in Environments plugin; fixed

TLDR; removes a redundant/faulty check for changes in the environments plugin

## Before:
Before this change there was a single unit test for environments. That test modelled 7 environments with a change in each.

image

Two of the seven changes (wait_timer, prevent_self_review) were detected by mergeDeep.compareDeep(); 5 were not detected.

image

The two changes that were detected were enough to lift the code execution past this point, deeper into the environments code which itself again uses it's own comparator() where this time it successfully finds the 7 changes.

image

It was theorized that if the unit test was decomposed into it's seven constituent pieces that only 2/7 would pass. This turned out to be true and is a better match to our lived experience with using environments in safe-settings:

image

Now on the theory that the first round of change detection was redundant, it was removed. This was accomplished by copying the sync method from diffable.js into environments.js and then removing the lines which failed to detect the changes. A new run of the unit tests, post change, passes 7/7:

image

But, what if the removed code wasn't really redundant? What if the existing config, and the desired config were the same; would environments.js now erroneously call GitHub to make a change when none was required? Nope:

image

## After:
The original test with 7 environments and 7 changes bundled together was added back in alongside the decomposed tests. It also passes. Here are the final results:

image

## Request:
I would be most appreciative of a review and merge of this change.  I am available for any follow up questions which may arise.


lib/plugins/environments.js
```diff
@@ -272,4 +273,79 @@ module.exports = class Environments extends Diffable {
...
+            return Promise.resolve(resArray)
+          }
+          return Promise.all(changes)
+        }).catch(e => {
+          if (this.nop) {
+            if (e.status === 404) {
+              // Ignore 404s which can happen in dry-run as the repo may not exist.
+              return Promise.resolve(resArray)
+            } else {
+              resArray.push(new NopCommand(this.constructor.name, this.repo, null, `error ${e} in ${this.constructor.name} for repo: ${JSON.stringify(this.repo)} entries ${JSON.stringify(this.entries)}`, 'ERROR'))
```
klutchell, line 342: 'NopCommand' is not defined here

klutchell:
I'm going to give this PR a try, with some minor changes from https://github.com/github/safe-settings/pull/616 on top.

lib/plugins/environments.js
Brad-Abrams: Thanks @klutchell .  I have added a pair of commits to address this: [cf3fa72](https://github.com/github/safe-settings/pull/646/commits/cf3fa7208e4123c59eeb379e18d4d179fca10ca3) and [caa7d04](https://github.com/github/safe-settings/pull/646/commits/caa7d04041f78632575a36abfee4f4e63c9ba841)
With these additions, the NopCommand is now defined.  Additionally, there was an argument missing from the constructor for the environments plugin when instantiated from the unit tests.  That has also been corrected.

I appreciate that you're taking a look at the PR.  I have more to follow on from this one.

klutchell: Any improvements to the environments plugin at this point is appreciated, as we are struggling to use this project to set environment level variables.

With PR https://github.com/github/safe-settings/pull/616 we were able to get it to create the initial set of environments and the respective variables, but now we can't get it to detect changes to those variables in the configuration.

Brad-Abrams: My experience suggests that this PR #646 is going to help with the detection and setting of environment level variables.  That being said, there are remaining quirks in the environments plugin that I'd still like to address.  For one, I can say that I appreciate the defensive code that you've added in PR# 616 to only post updates for variables and deployment protection rules if they are defined.  My plan has been to see if I can succeed in getting this succinctly scoped PR reviewed and merged before layering in additional improvements.  Am I on the right track?  Which one of us is going to take a swing at combining 616 and 646?  How would you feel about me taking it on? ... or would the resulting PR be too bloated?

klutchell: I'm not the project maintainer, but that sounds like the correct approach to me.

WDYT @decyjphr ?

Brad-Abrams:
I have a new PR inbound that will combine together this PR #646 and Kyle's PR #616.  It just needs to navigate some internal processes first.  It will additionally address issue #623 with a documentation update.

I also noticed that PR #630 relates to environments, however it has no material code change and so I've not taken it into consideration. ... I read it some more, and now I've taken it into consideration ... PR #630 stated that it was intended to address issue #628.  I'm going to see what I can do about #628

Brad-Abrams:
A new PR #649 supersedes this one, and combines in the changes from PR #616 as well.   

I did look into PR #630 which mentions environments but has no material code changes, and the underlying issue #628.  I may come back to it, however its of a lower priority and not included here.

luvsaxena1:
@Brad-Abrams  @klutchell We also found issue with adding or modifying environment variables second time once they are created. Looks like issue is with mergeDeep.js logic. I am also working to fix it. However, I have created issue for the same. 
Also looks like mergeDeep is currently broken not only for environments but also for Repository rulesets, adding or modifying bypass does not work or other fields to an existing rulesets does not work.

[Issue](https://github.com/github/safe-settings/issues/660)

klutchell:
@luvsaxena1 that issue aligns with what we are seeing, both changed variables and updated rulesets are not detected by mergeDeep.

luvsaxena1:
@klutchell  @Brad-Abrams  I have created [PR](https://github.com/github/safe-settings/pull/665) to address merge deep code issues. I believe changes, I made, will successfully detect changes to the existing environment variables and other settings.

Brad-Abrams:
With PR #649 having been approved and merged, this PR is now redundant.  Closing.

PR #648, opened 2024-06-20
import eta templates as javascript strings

Oatelaus:
This pull request aims to move away from importing eta templates using the file system location and instead rely on the node module resolution system, passing the template as a string.

The underlying reason for this change is because the current implementation will break bundled deployments which are fairly common within a serverless deployment and I cannot see negatives to non-bundled users of this package.


decyjphr:
Awesome!  Thanks @Oatelaus

PR #669, opened 2024-08-27
fix code scanning alerts

### LINKED ISSUE #668: Fix code scanning alerts

decyjphr:
Fixes #668

### DIFF
--- a/lib/glob.js
+++ b/lib/glob.js
@@ -1,40 +1,42 @@
+const _ = require('lodash')
 class Glob {
-  constructor(glob) {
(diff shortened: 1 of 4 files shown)


luvsaxena1:
@decyjphr  These changes are making safe setting to malfunction and are implementing settings to the repos which are not configured via safe settings by reading from some file (in our case it picked one) from the suborg configuration. 

It has impacted 81 repos in our org. We have reverted these changes in our local version of safe settings for now. We had plans to invest time in fixing this. But just letting you know that this is not working as expected.

PR #670, opened 2024-08-28
chore: batch dependency updates from dependabot

jmeridth:
To minimize the number of pull requests, this PR batches dependabot dependency update pull requests.

When minor/patch changes to dependencies, they are batched together in a pull request.

Major dependency changes are still their own PR to ensure visibility.


jmeridth:
@decyjphr what do you think about this?  Or do we prefer the split dependency PRs for visibility?  Each project operates its own way.  Curious what you think?

PR #676, opened 2024-09-05
Support uppercase environment variable names

klutchell:
Github REST API supports case-sensitive names on this endpoint
so we should always create (POST) variables with the desired case.

On subsequent GET, PATCH, and DELETE requests the Github REST API
will use loose matching to find any existing objects, ignoring case.

So in general we need to force to lowercase to perform diffs, and respect
the provided case for create, but for all other calls it doesn't matter
which case is used.


klutchell:
Not required if https://github.com/github/safe-settings/pull/682 is merged

PR #682, opened 2024-09-14
Fix: array comparison breaks if using the environments plugin

gramatus:
I believe this will solve what I think is a serious bug introduced in version 2.1.11 as reported in issue 108.

If I understand things correctly, loading the environment plugin removes the `name` value from the `MergeDeep.NAME_FIELDS` array and this again breaks merging arrays, thus leading to diff reporting not working as expected.

I believe this change keeps the result of the fix that caused the bug, but in a way that removes the problem.

I also did a quick change at the end of `mergeDeep.js` where this line was duplicated (I believe that must be just a slip):
```javascript
MergeDeep.NAME_FIELDS = NAME_FIELDS
MergeDeep.NAME_FIELDS = NAME_FIELDS
module.exports = MergeDeep
```

### DIFF
--- a/lib/mergeDeep.js
+++ b/lib/mergeDeep.js
@@ -378,5 +378,4 @@ class MergeDeep {
   }
 }
 MergeDeep.NAME_FIELDS = NAME_FIELDS
-MergeDeep.NAME_FIELDS = NAME_FIELDS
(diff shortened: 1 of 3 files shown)


decyjphr (approved):
Looks good to me

PR #664, opened 2024-08-08
Same repository that is part of multiple suborgs

decyjphr:
Added code for the following items:
- [x] Refactoring Sub-Organization Configurations: Throw an error if a repo is found for multiple sub-organizations. Fixes #277 
- [x] Unit Test Enhancements: Added tests for loading and checking suborg configs
- [x] ESLint config changes

PR #683, opened 2024-09-15
Updated according to lint rules

gramatus-old:
This PR makes two changes to `.eslintrc.json` and then updates the remaining code to follow the current linting rules.

These are the changes that are not pure updates to follow linting standards:
1. Change `parserOptions.ecmaVersion` from 12 to 13. This is due to code in this project that fails the lint test in ecma 12.
2. Change from ignoring `test/**/*.js` to overriding the `env` to `jest` for these files. By doing this we can get linter hints in the editor without the linter failing.

All other changes are either done by running `npx standard --fix`, or some manual renaming of variables from snake_case to camelCase, again to adhere to project standards.

Feel free to discard this PR, but I am hoping it is a useful improvement to the codebase.


gramatus-old:
@decyjphr I see that the eslint config is changing from extending "standard" to extending "eslint:recommended" in #664. Would it be interesting if I made a new version of this, but using the new Eslint rules? I am hoping the repo can reach a the point where there are no lint errors, thus allowing enforcing linting on PRs, and would be happy to contribute towards that goal if it aligns with your goals.

PR #684, opened 2024-09-15
Updated tests

gramatus-old:
This is so far a WiP, but I am trying to see if I can get those tests that are currently skipped to work.

I have managed to run all the tests in `index.test.js`, `branches.test.js` and `milestones.test.js`, but I am not completely certain that all the tests actually checks what they should. I will try to look into that later. I believe this is still a good way towards better testing.

One important thing to note is that I added `title` to `MergeDeep.NAME_FIELDS` in order to make milestones work. I am not sure if this might have unintended consequences, and would be happy if someone else has any insights into that.

I also added a few JSDoc types (for my own understanding of the code), hopefully that is not a problem.

Finally, I fixed a few places where I believe the logging was wrong (see 6f205c3).


gramatus-old:
@decyjphr does this look promising? It might be a few weeks before I can continue, but if this is interesting I will try to make sure all the tests in this PR works and does what is expected.

PR #689, opened 2024-09-20
feat: add JSON schema doc for settings

jonathanmorley:
There is (at least) an extra `security_and_analysis` key in the `repositories` section, that I don't think would get correctly handled, but, this should be a superset of allowable configuration.


decyjphr:
This PR is merged after reverse merging from main.

Issue #559, opened 2023-11-18
Rate limit issues 


bheemvennapureddy:
Hitting rate limits with Cron running every min 

<img width="877" alt="Screenshot 2023-11-17 at 10 24 13 PM" src="https://github.com/github/safe-settings/assets/144472943/6864a821-a191-4932-b4cf-f308e0963692">

bheemvennapureddy:
<img width="1217" alt="Screenshot 2023-11-17 at 10 30 30 PM" src="https://github.com/github/safe-settings/assets/144472943/832a5f23-e2f0-4607-9a02-2a27141daa1a">

svg153:
The CRON is not designed to be launched every minute. It is a reconciliation system that was put in place to consolidate the state in the safe-settings files and the real state of github, with the aim of not having a rate limit with so many requests to the API. Why do you have Chrome every minute? What are you trying to solve?
cc: @decyjphr

bheemvennapureddy:
@svg153 i assume CRON is to gurantee that all the policies are correctly applied and some one trying to overwrite any of these policies like say adding labels manually should be overwritten by Safesettings no ? same goes to any status checks or branch protection rules cc: @decyjphr 

i feel there is a bug in the system which is causing the rate limit issues - some stats below what i received from github 

<img width="1496" alt="Screenshot 2023-11-21 at 7 46 53 PM" src="https://github.com/github/safe-settings/assets/144472943/a675d660-ebf7-4cf0-ba58-99f0cd555546">

[Top-url-CuroFinTechCorp.csv](https://github.com/github/safe-settings/files/13432638/Top-url-CuroFinTechCorp.csv)
[Top_route-CuroFinTechCorp.csv](https://github.com/github/safe-settings/files/13432641/Top_route-CuroFinTechCorp.csv)

martinm82:
What is the number of repositories within your organization? We have the same issue with rate limits but in our case the issue is that safe-settings just does not scale with large orgs (in our case >10k repos). We needed to change the cycle of the cron to be executed every 4-6 hours as otherwise we would hit immediately the rate limit.

Some changes, like branch protection rules are almost immediately changed back to the ones defined by the safe-settings configurations. This is achieved by webhook events (see index.js for the events that trigger an update of the repo settings).

[...]

PR #692, opened 2024-09-25
feat: add unarchiving plugin

### LINKED ISSUE #443: [Feature Request] Skip archived repositories from being configured
## New Feature

When configuring (e.g., topics, vulnerability, general settings, etc. ) archived repositories, safe-settings receives HTTP 422 errors. Since the intention of archived repositories is to not use them anymore there is no reason to configure them.

Therefore my proposal is to exclude those in safe-settings. This will have a positive impact on the rate limit.

randomcascade:
Meant to resolve #443 may help #559 

Inspired by #193 I've added a one-off plugin. It is called archive. In most cases it does nothing, but in case a config for archiving a repo is true and the real repo is not archived this plugin will archive that repo and prevent other plugins from running. This does mean config may not be provided to an archived repo, but since it's being archived that hardly matters. This will also clobber the diff comment for PRs where archived is set to true. We may try setting the nop flag to true in the future if the repo is archived by the Archive plugin, but for the purposes of avoiding the rate limit this should be fine.


randomcascade:
The unit test is rudimentary. Running locally for many cycles has not caused issues, but more tests or review would be appreciated.

Alex-Mills-seccl:
👍 this would be super useful for me

HagegeR:
Hi, I'm not sure if I should open an issue on this, but when trying the release related to this PR, I get a lot of unnecessary calls to repo.change_merge_setting in my use case, I think it's adding empty changes to the repository config
I'm using full-sync from a GitHub Workflow if that can help narrow down the issue root cause

PR #708, opened 2024-12-11
fix: Handle scientific notation in label hex colors

### LINKED ISSUE #707: Label color parsing fails with scientific notation hex values
## Problem Description

When label colors are specified without quotes in the YAML they could be interpreted as scientific notation (e.g. `5319e7`), they are incorrectly converted to decimal number

```yaml
labels:
  - name: "example"
    color: 5319e7  # 53190000000 instead of 5319e7
```

GitHub API return status code 422 since value is not a valid.

The label creation/update fails

Unquoted color values should be preserved as hex colors rather than being converted to decimal notation

```json
{
  "message": "Validation Failed",
  "errors": [{ "resource": "Label", "code": "invalid", "field": "color" }]
}
```

PendaGTP:
Fixes #707

Properly convert numeric hex colors that could be interpreted as scientific notation (e.g. 5319e7)

## Before
```yml
labels:
  - name: "example"
    color: 5319e7  # 53190000000
```

## After
```yml
labels:
  - name: "example"
    color: 5319e7  # 5319e7
```

### DIFF
--- a/lib/plugins/labels.js
+++ b/lib/plugins/labels.js
@@ -23,7 +23,13 @@ module.exports = class Labels extends Diffable {
       this.entries.forEach(label => {
         // Force color to string since some hex colors can be numerical (e.g. 999999)
         if (label.color) {
-          label.color = String(label.color).replace(/^#/, '')
+          // Handle scientific notation case for numerical color values
+          label.color = (
+            typeof label.color === 'number'
+              ? label.color.toString(16)
+              : String(label.color)
+          ).replace(/^#/, "")
+          label.color = String(label.color).replace(/^#/, "")
           if (label.color.length < 6) {
             label.color.padStart(6, '0')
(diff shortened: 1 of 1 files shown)


anderssonjohan:
@PendaGTP Please provide a test for your code. Such as:

```javascript
const actualString = parseColorFromScientificNumber(5319e7);

expect(actualString).toEqual('5319e7');
```

Do you get these results with the provided code?

PendaGTP:
Thanks for your feedback. I apologize my local testing was wrong.

The Issue stems from `yaml.load` converting scientific notation to its decimal. After consideration, I don't believe this warrants implementing a custom schema for the color, especially since the doc already specifies that labels should be quoted.

We can close the PR and the issue related.

I appreciate your time reviewing. Thanks you.

PR #710, opened 2024-12-11
fix: create new repositories during full-sync operation

### LINKED ISSUE #709: New repositories not created during full-sync operation
## Problem Description

When running safe-settings via GitHub Actions using the full-sync script, new repositories defined in the configuration are not being created, while all other settings synchronization works as expected.

## Current vs Expected Behavior

**Currently:**

- Full-sync runs successfully without errors
- Existing repositories are properly synchronized
- New repositories defined in config are not created

**Expected:**

- New repositories should be created based on the configuration during full-sync

## Technical Analysis

The issue stems from different code execution paths:

1. **Normal Operation (Working via webhook)**

   - Event triggers `syncSettings` → `Settings.sync` → `Settings.updateRepos`
   - Directly processes repositories without filtering

2. **Full-Sync (Not Working)**
   - Process: `syncInstallation` → `syncAllSettings` → `Settings.syncAll` → `Settings.updateAll()`
   - `updateAll` only processes existing repositories via `eachRepositoryRepos` (which use `GET /installation/repositories` to filter)
   - Missing: Check for repositories defined in config but not yet created

## Environment

- safe-settings: 2.1.24
- Platform: github.com
- Running on: Github Action (cron + manual trigger)
- Using full-sync script

## Related Issues

- #460

PendaGTP:
Fixes #709 

The full-sync script now properly creates new repositories defined in the configuration instead of only updating existing ones. This fixes the different behavior between webhook-based sync and full-sync runs.


lib/settings.js
```diff
@@ -475,19 +475,50 @@ ${this.results.reduce((x, y) => {
...
+      .map(config => {
+        return {
+          name: config.repository.name,
+          owner: config.repository.organization || this.repo.owner
+        }
+      })
+    const missingRepoResults = await Promise.all(
+      repoInConfigs
+        .filter(repo => {
+          return !processedRepos.has(`${repo.owner}/${repo.name}`) || this.isRestricted(repo.name)
```
decyjphr: @PendaGTP this logic seems allow restricted repos because only time the `this.isRestricted` is checked only of a repo is processed. even the that logic seems to be incorrect.
```suggestion
          return !(processedRepos.has(`${repo.owner}/${repo.name}`) || this.isRestricted(repo.name))
          // This would be same as 
          // return !processedRepos.has(`${repo.owner}/${repo.name}`) && !this.isRestricted(repo.name)
```

decyjphr (changes requested):

lib/settings.js
PendaGTP: @decyjphr Thanks for your review. I've pushed an update that fixes this wrong logic.

PR #716, opened 2024-12-20
CONFIG_REPO does not exist, change to correct ADMIN_REPO

karl-petter:
Rename one variable in the documentation to match the existing one


decyjphr (approved):
Thanks for correcting.

PR #731, opened 2025-01-06
fix: handle empty deployment config yaml file

### LINKED ISSUE #730: Empty deployment-settings.yml cause error
## Problem Description

Currently safe-settings doesn't handle empty `deployment-settings.yml` and null arrays safely.
This cause runtime errors when the deployment config is loaded/processed.

**Proposed solution:**

- Add fallback for empty yaml
- Add array validation before iteration

## Environment

- safe-settings: 2.1.24

PendaGTP:
Fixes #730

Changes:
- Add `|| {}` fallback for yaml.load() results
- Add `|| []` fallback for array properties (`overridevalidators` and `configvalidators`)
- Add array validation before iteration

Testing:
- Verified handling of empty config file
- Verified handling of null array cases

### DIFF
--- a/lib/deploymentConfig.js
+++ b/lib/deploymentConfig.js
@@ -15,21 +15,21 @@ class DeploymentConfig {
   static {
     const deploymentConfigPath = process.env.DEPLOYMENT_CONFIG_FILE ? process.env.DEPLOYMENT_CONFIG_FILE : 'deployment-settings.yml'
     if (fs.existsSync(deploymentConfigPath)) {
-      this.config = yaml.load(fs.readFileSync(deploymentConfigPath))
+      this.config = yaml.load(fs.readFileSync(deploymentConfigPath)) || {}
     } else {
       this.config = { restrictedRepos: ['admin', '.github', 'safe-settings'] }
     }
 
-    const overridevalidators = this.config.overridevalidators
-    if (this.isIterable(overridevalidators)) {
+    const overridevalidators = this.config.overridevalidators || []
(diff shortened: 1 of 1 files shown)


PendaGTP:
Current implementation: `deployment-settings.yml` uses defaults when missing but not when empty.

Consider either:
- using defaults for both missing and empty files
- documenting that empty files clear defaults

Thoughts?

HagegeR:
> Current implementation: `deployment-settings.yml` uses defaults when missing but not when empty.
> 
> Consider either:
> 
> * using defaults for both missing and empty files
> * documenting that empty files clear defaults
> 
> Thoughts?

Shouldn't empty values skip the application of the rule (keep current setting on the repo)?

PendaGTP:
> > Current implementation: `deployment-settings.yml` uses defaults when missing but not when empty.
> > Consider either:
> > 
> > * using defaults for both missing and empty files
> > * documenting that empty files clear defaults
> > 
> > Thoughts?
> 
> Shouldn't empty values skip the application of the rule (keep current setting on the repo)?

To clarify, this change is specifically about `deployment-config.yml` (example [here](https://github.com/github/safe-settings/blob/main-enterprise/docs/sample-settings/sample-deployment-settings.yml)), which is the app's core configuration file. Unlike other settings files, this configuration cannot be overridden.

Regarding the original question about empty vs missing files:
This current implementation applies defaults when the file is missing, but not when the file is empty.

Please let me know if I've missed something or if you need additionnal clarification.

PR #732, opened 2025-01-07
refactor: standardize variable file naming and merge glob patterns

PendaGTP:
## Changes

_No functional changes_

- Extracted sub-org config pattern into `Settings.SUB_ORG_PATTERN`
- Extracted repo config pattern into `Settings.REPO_PATTERN`
- Rename `Settings.FILE_NAME` to `Settings.FILE_PATH` for better clarity and consistency
- Rename `env.DEPLOYMENT_CONFIG_FILE` to `env.DEPLOYMENT_CONFIG_FILE_PATH`
  - I didn't rename env var since this will cause a breaking change

Note: We could also consider to add env var `DEPLOYMENT_CONFIG_FILE_PATH` to support both old and new naming in the future. This would allow us to introduce it in a minor version while maintaining backward compatibility. Removing the old env vars would then be a breaking change for a future major version.

## Why? 

I'm currently deep dive into the code base to try fix some issues and doing other refactor to help us contribute more easily.
I will try to submit PR with some small changes instead of big one and no breaking change if possible.
Let me know if this approach works for you - I want to be helpful, not disruptive.

### DIFF
--- a/.github/workflows/create-pre-release.yml
+++ b/.github/workflows/create-pre-release.yml
@@ -50,7 +50,7 @@ jobs:
           cache: 'npm'
       - run: npm install
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@6524bf65af31da8d45b59e8c27de4bd072b392f5
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2
       - name: Log in to the Container registry
         uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567
         with:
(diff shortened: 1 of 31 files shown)


HagegeR:
I also felt this inconsistency when using safe-settings as a github action, I think if this gets merged it'll make the first time users life a lot easier

PendaGTP:
> I also felt this inconsistency when using safe-settings as a github action, I think if this gets merged it'll make the first time users life a lot easier

To clarify, this PR currently only affects the backend implementation and does not change any environment variables that users need to configure.

In the futur, I think it would be beneficial to also rename some exposed environment variables to make them more intuitive.

However, I believe it might be too early to implement theses changes now. I would be better to include them with other breaking changes (new major version) in a future release to avoid frustating users with frequent breaking changes. 
As an alternative, we could improve the documentation to make the current environment variable names and their purposes clearer to users. Personally, I find the current environment variable documentation clear enough, but if you have any suggestions for improvements, please feel free to share them.

This is just my current opinion, so open to other perpectives.

PR #735, opened 2025-01-08
feat(extension): also allow .yaml file format

### LINKED ISSUE #700: Yaml file extension: .yml or .yaml
## Prerequisites:

* Is the functionality available in the GitHub UI? 

No / N/A

* Is the functionality available through the GitHub API? 

No / N/A

## New Feature

The [README](https://github.com/github/safe-settings?tab=readme-ov-file#github-safe-settings) states that files with the `.yaml` file extension will be ignored. Yet both `.yaml` and `.yml` are valid. Therefore, I've got a couple of suggestions:

a. Either return an error when the `.yaml` file extension is in use, in order to notify users they should switch to `.yml`.
b. Support both file extensions. Could it be around [here](https://github.com/github/safe-settings/blob/777a97f489621bbff84eaf1dd655f10ee84e2942/lib/settings.js#L645)?

dabcoder:
Fixes https://github.com/github/safe-settings/issues/700.


dabcoder:
cc @anderssonjohan and @decyjphr if you have time to have a look

Issue #719, opened 2024-12-25
Doing schema validation for safe-settings.


decyjphr:
@jonathanmorley had raised a PR #689  to merge JSON schema for validating `settings.yml`. This is a great start! And  @luvsaxena1, @zpratt, @primetheus, have also  done various implementations of schema validation checks in their environments to catch errors in the settings yaml -- in the IDE or during the PR check.  

I am new to this and I've been looking through the docs for [Zod](https://zod.dev/) (Which Zach has been using) and [AJV](https://ajv.js.org/) (Which I found via search). 

I created this issue to allow us to collaborate and take their work and have some sort of schema validation in safe-settings.

primetheus:
@decyjphr I looked at `ajv`, but I haven't seen `zod`. Ours is built using [openapi-request-validator](https://www.npmjs.com/package/openapi-request-validator), but we're hitting an issue with the `rulesets` due to an issue with traversing `$ref` in the `enforcement` in [@octokit/openapi](https://www.npmjs.com/package/@octokit/openapi) schemas. I'm definitely interested in seeing how others are handling rulesets

PR #738, opened 2025-01-10
fix: add missing nop logic for custom properties plugin

### LINKED ISSUE #734: Custom properties are applied during pull request
## Problem Description
When in PR mode, adding a custom property results in the property added before being merged into main.

Can see the changes in the PR, once i did a different unrelated change, i could see autolinks still waiting to be applied but the custom properties were applied.

Custom properties are only applied in main 

```

```

Running our own in PR mode
Github.com
latest

PendaGTP:
Fixes #734

Implements proper `nop` (dry-run) logic handling for Custom PropertIes plugin, aligning with documentation and other plugins.

## Changes

- Add nop mode support for property modifications
- Add `paginate` for gettings all custom properties
- Switch to built-in Octokit functions instead of `github.request`
- Update API endpoint format from `/repos/:org/:repo/properties/values` to `/repos/{owner}/{repo}/properties/values` for consistency with GitHub API
- Refactor property modification logic into a single `modifyProperty` method
- Add debug logging for better troubleshooting
- Normalize custom property names (from GH API) to lowercase to prevent case-sensitivity comparison issues
- Update test cases

## Note

Currently, organization properties are not fetched from the GH Organization API. When modifying repo custom properties, this results in a 422 error if the property isn't defined at the org level. Future improvements could:

- Handle missing org properties to enable preview changes in nop mode
- Add support for defining Organization custom properties (requires broader changes)

### DIFF
--- a/lib/plugins/custom_properties.js
+++ b/lib/plugins/custom_properties.js
@@ -1,25 +1,47 @@
 const Diffable = require('./diffable')
+const NopCommand = require('../nopcommand')
 
 module.exports = class CustomProperties extends Diffable {
   constructor (...args) {
     super(...args)
 
     if (this.entries) {
-      // Force all names to lowercase to avoid comparison issues.
-      this.entries.forEach(prop => {
-        prop.name = prop.name.toLowerCase()
-      })
+      this.normalizeEntries()
     }
   }
 
-  async find () {
-    const data = await this.github.request('GET /repos/:org/:repo/properties/values', {
-      org: this.repo.owner,
-      repo: this.repo.repo
-    })
+  // Force all names to lowercase to avoid comparison issues.
+  normalizeEntries () {
+    this.entries = this.entries.map(({ name, value }) => ({
+      name: name.toLowerCase(),
+      value
+    }))
+  }
 
-    const properties = data.data.map(d => { return { name: d.property_name, value: d.value } })
-    return properties
+  async find () {
+    const { owner, repo } = this.repo
+    const repoFullName = `${owner}/${repo}`
+
+    this.log.debug(`Getting all custom properties for the repo ${repoFullName}`)
+
+    const customProperties = await this.github.paginate(
+      this.github.repos.getCustomPropertiesValues,
+      {
+        owner,
+        repo,
+        per_page: 100
+      }
+    )
+    this.log.debug(`Found ${customProperties.length} custom properties`)
+    return this.normalize(customProperties)
+  }
+
+  // Force all names to lowercase to avoid comparison issues.
+  normalize (properties) {
+    return properties.map(({ property_name: propertyName, value }) => ({
+      name: propertyName.toLowerCase(),
+      value
+    }))
   }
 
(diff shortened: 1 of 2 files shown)


wernerb:
It has a bug where in the PR it states:
```
error TypeError: Cannot read properties of undefined (reading 'toLowerCase') in CustomProperties for repo: {\\\"owner\\\":\\\"x\\\",\\\"repo\\\":\\\"y\\\"} entries [{\\\"name\\\":\\\"type\\\",\\\"value\\\":\\\"myvalue\\\"}]\",
```

PendaGTP:
> It has a bug where in the PR it states:
> 
> ```
> error TypeError: Cannot read properties of undefined (reading 'toLowerCase') in CustomProperties for repo: {\\\"owner\\\":\\\"x\\\",\\\"repo\\\":\\\"y\\\"} entries [{\\\"name\\\":\\\"type\\\",\\\"value\\\":\\\"myvalue\\\"}]\",
> ```

@wernerb I have not been able to reproduce this issue. Cloud you please share your config files?

I've tried the following scenarios:
- add/remove/update `custom_properties` in `settings.yml`
- add/remove/update `custom_properties` in `repos/repo.yml`
- adding and remove org custom properties through the UI
- updatating repository custom properties values through the UI

Until a more comprehensive solution is implemented (#719), we could consider adding basic validation:
- Validate that entries is an array
- Verify each array item is an object with a defined name property of type string (and value must be null/undefined or type of string)

Thanks!

wernerb:
@PendaGTP I use the official docker image 2.15-rc.1 as there is a problem with the newer images (see latest issue where i get a 404 on using the image). I override `custom_properties.js` and mount that in the container. In a suborg I modify a custom property and the error happens and is seen in the PR. If you need help reproducing feel free to hit me up on github in DMs. Unfortunately this is high priority for us. 

It is such high-priority, that I am tempted to do away with our hosted flow and use the action flow instead as it seems it is more supported?

The custom property is a "text" type property nothing really special with it

PendaGTP:
@wernerb see https://github.com/github/safe-settings/issues/743#issuecomment-2607189122

PendaGTP:
Additionally, there are currently reported issues with the full-sync script used in GitHub Actions (GHA), so I recommend against migrating to it at this time.
Related issues: #739 #378 #709

HagegeR:
> It is such high-priority, that I am tempted to do away with our hosted flow and use the action flow instead as it seems it is more supported?

what a coincidence, I was thinking of doing exactly the inverse and use the hosted flow instead 😀

grossag:
I am really interested in this 😄. My company is using safe-settings but I don't want to start deploying custom properties until we get no-op support.

As a clarifying question, are we still able to manage some custom properties manually and some custom properties through safe-settings? I know for other things (e.g. topics, teams, and people), once we start managing them in safe-settings we can't manage them manually. But I am not sure for custom properties if it's an all-or-nothing thing or if it is a one-by-one for custom properties.

PR #750, opened 2025-01-24
fix!: replace custom glob with minimatch for improved pattern matching

### LINKED ISSUE #748: `suborgproperties` is matching too many repositories
## Problem Description

I have a suborg file:
```
suborgproperties:
  - sub-org: horizon-cart

repository:
  topics:
    - horizon-cart
```

This rule matches 11 repositories in my organization, one of which is `euc-vdi/cart`. So we end up with this entry in `settings.js` `this.subOrgConfigs`:
```
"cart": {
  "suborgproperties": [{"sub-org":"horizon-cart"}],
  "repository":{"topics":["horizon-cart"]},
  "source":".github/suborgs/horizon-cart.yml"
}
```
That is correct, but this function implementation does too loose of matching:
```
  getSubOrgConfig(repoName) {
    if (this.subOrgConfigs) {
      for (const k of Object.keys(this.subOrgConfigs)) {
        const repoPattern = new Glob(k)
        if (repoName.search(repoPattern) >= 0) {
          return this.subOrgConfigs[k]
        }
      }
    }
    return undefined
  }
```
It causes any repository name starting with `cart` to match this rule!

Can this be fixed by changing the function to simply be this:
```
  getSubOrgConfig(repoName) {
    return this.subOrgConfigs[repoName]
  }
```

Running our own.

GitHub Enterprise Cloud, so github.com

This issue happens when deploying either https://github.com/github/safe-settings/commit/a692dbf506782c16fcf1cb995548cc6f74ced02a or https://github.com/github/safe-settings/commit/774b2e5d64a9f27a9dfadcc1afe6be944eece375

Using cloud.

PendaGTP:
### Description

Fixes #748 

Replace custom glob implementation with [minimatch](https://github.com/isaacs/minimatch).

### Changes

- Replace custom glob implementation with `minimatch` to fully support glob patterns
- Remove support for Regex patterns
- Simplify logic for matching glob patterns
- Update and improve related documentation

:warning:  **BREAKING CHANGES:** :warning: 

Removed Regex support for patterns. Users must update their configuration to use glob syntax.

Before:

```yml
exclude: ['^admin$', '^\.github$', '^safe-settings$', '.*-test']
```

After:

```yml
restrictedRepos: ['admin', '.github', 'safe-settings', '*-test']
```

This change simplifies the configuration and doesn't significantly impact existing functionality.

### Testing

- Ensured all unit tests pass
- Removed glob.test.ts as it's no longer necessary
- Performed minimal manual testing


PendaGTP:
Note:

This PR doesn't impact:
- labels regex
- validator regex

decyjphr:
This is great! But requires some manual resolution of conflicts. So will merge it after handling a few simpler PR merges.

PendaGTP:
Let me know if (and when) you need help to resolve conflicts.

decyjphr:
Thanks @PendaGTP if you could look at that it would be awesome. It is mostly the code you have contibuted in the previous PRs and this, so it would be easier for you.

PendaGTP:
@decyjphr fyi rebase done

bmike78:
I do want to comment that this fix by @PendaGTP works much better than the recent #808 merge when it comes to mistaken globbing issues.

Basically as it stands with current ```main-enterprise``` branch with #808, if there is a config like this:

```
suborgrepos:
  - luv
```

Then it will be greedy and pull in any repos that match "luv" in any part of the repo name, whereas in this PR, the behavior is that this config would only match the repo called "luv" and that's all.

## Testing

* Tested globbing works, for example:

```
suborgrepos:
  - luv*
```

This will match any repo that starts with "luv", for example: "luv-mike-repo" will match.

* This did not seem to work:

```
suborgrepos:
  - *luv*
```

Our team is good with this behavior, just pointing out that we also tested this scenario.

## Background

* About 7-8 months ago, we had a subort config with "admin" in the name that caused safe settings to deploy on several dozen repos that matched "admin" in any part of the repo name.
* Due to this, we diverged from the standard glob.js, but it looks like this PR will allow us to join back up with upstream safe settings once this PR is merged.

decyjphr:
@PendaGTP wondering if you have seen this error before
```
ERROR (event): Cannot read properties of undefined (reading 'split')
```
Trying to test this before merging it.

index.js
```diff
@@ -134,75 +134,60 @@ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) =>
...
+
+    const changes = [
+      ...getMatchingFiles(payload.commits, 'added'),
+      ...getMatchingFiles(payload.commits, 'modified')
+    ]
+
+    return changes.map((file) => ({
+      name: file.match(/([^/]+)\.yml$/)[1],
+      path: file
+    }))
```
decyjphr: The returned object must have `repo` property similar to context.repo() in order for this to work. Otherwise it will throw `Cannot read properties of undefined (reading 'split')` error when processing in settings.js
```suggestion
    return changes.map((file) => ({
      repo: file.match(/([^/]+)\.yml$/)[1],
      path: file
    }))
```

decyjphr (changes requested):

decyjphr (approved):
LGTM

PR #761, opened 2025-02-18
Unsafe Settings

jonathanmorley:
This adds the capability for safe-settings to treat some fields as 'unsafe', and allow them to be configured by the individual repositories.

### DIFF
--- a/lib/deploymentConfig.js
+++ b/lib/deploymentConfig.js
@@ -11,6 +11,7 @@ class DeploymentConfig {
   // static config
   static configvalidators = {}
   static overridevalidators = {}
+  static unsafeFields = []
 
   static {
     const deploymentConfigPath = process.env.DEPLOYMENT_CONFIG_FILE ? process.env.DEPLOYMENT_CONFIG_FILE : 'deployment-settings.yml'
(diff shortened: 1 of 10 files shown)


HagegeR:
could you update the read-me to reflect how to use this feature?

jonathanmorley:
Because the settings and deploymentConfig objects are combined to a single object, the organization settings can override/set values that are intended to come from the deploymentConfig.

This PR separates them, as this was felt to be _too_ unsafe for this deploymentConfig setting

jonathanmorley:
This should get rebased on top of https://github.com/github/safe-settings/pull/768.

lib/deploymentConfig.js
```diff
@@ -36,6 +37,15 @@ class DeploymentConfig {
         this.configvalidators[validator.plugin] = { isValid: f, error: validator.error }
       }
     }
+
+    const unsafeFields = this.config.unsafeFields
+    if (unsafeFields) {
+      if (this.isIterable(unsafeFields)) {
```
decyjphr, line 43: this.isIterable(...) is not a method on this class. It used to be there but is not in this branch

decyjphr (changes requested):

PR #783, opened 2025-03-07
filter installations by GH_ORG

### LINKED ISSUE #782: don't take fixed: installations[0]
## Problem Description

Running as GHA, with multiple apps being installed (in different orgs), but not all have the same ADMIN_REPO incl. the settings, let the app fail.

We had a scenario, where one app was installed in multiple different orgs (one enterprise account and two personal accounts). When the app is now executed as [GHA](https://github.com/github/safe-settings/blob/main-enterprise/docs/github-action.md), it is checking for its installations and [always just uses the first occurrence, when there are more than one](https://github.com/github/safe-settings/blob/9dbe6c72ba9066c86547a0a924e43287b8a5d761/index.js#L242).

Do NOT simply use the first installation, but instead use the one, which matches the GH_ORG environment setting.

```text
DEBUG (github): GitHub request: GET https://api.github.com/repos/<private-account>/.github/contents/safe-settings%2Fsettings.yml - 404
ERROR (probot): Error getting settings HttpError: Not Found
Unexpected error during full sync: TypeError: Cannot read properties of undefined (reading 'data')
```

running your own as described in https://github.com/github/safe-settings/blob/main-enterprise/docs/github-action.md

on github.com

2.1.16

hilmarf:
fixes: https://github.com/github/safe-settings/issues/782

### DIFF
--- a/index.js
+++ b/index.js
@@ -221,12 +221,14 @@ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) =>
       github.apps.listInstallations.endpoint.merge({ per_page: 100 })
     )
     robot.log.debug(`installations: ${JSON.stringify(installations)}`)
-    if (installations.length > 0) {
-      const installation = installations[0]
+    const installation = installations.find(inst => inst.account.login === env.GH_ORG)
+    if (installation) {
       const github = await robot.auth(installation.id)
       const app = await github.apps.getAuthenticated()
       appSlug = app.data.slug
       robot.log.debug(`Validated the app is configured properly = \n${JSON.stringify(app.data, null, 2)}`)
+    } else {
+      robot.log.error(`No installation found for organization: ${env.GH_ORG}`)
     }
   }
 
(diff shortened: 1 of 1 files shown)


decyjphr:
The env variable `GH_ORG` is a probot [configuration](https://github.com/probot/probot/blob/master/docs/configuration.md) for `manifest` flow. That is not a required env variable so in most cases. it would not be set. 

@hilmarf there might a different way to get the org in question. Let me investigate.

Simon-Boyer:
@decyjphr  Could we simply check if that variable (or even another one we define) is there and use it, if not we revert back to using installations[0]?

renan-alm:
Right now this is a limitation. 
This feature is needed for managing which orgs are in scope of the settings. Otherwise, we (users) are limited to use a single private GH App per org. In other words, it is a low-handing fruit. 

The only work left on this PR is adding tests.

PR #792, opened 2025-03-21
Fix a bug in fetching sub org repos by custom properties

decyjphr:
There was a bug introduced during refactorings in the `lib/settings.js` file which resulted in sub-org repos not being fetched correctly for custom properties.

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -321,61 +321,48 @@ ${this.results.reduce((x, y) => {
       }
     }
 
-        // Overlay repo config
+    // Overlay repo config
     // RepoConfigs should be preloaded but checking anyway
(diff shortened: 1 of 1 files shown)


decyjphr:
@PendaGTP I had to make a slight modification in the code for fetching repos by suborg properties because it was failing with the way it was fetching previously. Since the change involved the octokit query, I reset it back to the way it was where it was querying the organization for all repos matching the custom property value.

PendaGTP:
@decyjphr I confirm there was an issue with searching repositories by custom property. GH docs indicate we should use `props.PROPERTY:VALUE` syntax instead of `props.PROPERTY.VALUE`. Sorry for missing this.

decyjphr:
No worries I wanted to keep you in the loop. I didn't catch it either when I merged it. Thanks @PendaGTP

PR #796, opened 2025-03-26
feat: Add support for tags in deployment branch policies

JakubBiel:
1. Add support for `tag` type in environments' deployment branch policies - see https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28#create-a-deployment-branch-policy
2. Update docs.
3. Update unit tests for environments.

### DIFF
--- a/lib/plugins/environments.js
+++ b/lib/plugins/environments.js
@@ -14,6 +14,16 @@ module.exports = class Environments extends Diffable {
             variable.name = variable.name.toLowerCase()
           })
         }
+        // Expand custom_branch_policies to match GitHub schema.
+        if (environment.deployment_branch_policy && environment.deployment_branch_policy.custom_branch_policies) {
+          const policies = []
+          environment.deployment_branch_policy.custom_branch_policies.forEach(policy => {
+            policy.names.forEach(name => {
+              policies.push({ name: name, type: policy.type })
+            })
+          })
+          environment.deployment_branch_policy.custom_branch_policies = policies
+        }
       })
     }
   }
@@ -52,7 +62,8 @@ module.exports = class Environments extends Diffable {
                 repo: this.repo.repo,
                 environment_name: environment.name
               })).data.branch_policies.map(policy => ({
-                name: policy.name
+                name: policy.name,
+                type: policy.type
               }))
             },
         variables: (await this.github.request('GET /repos/:org/:repo/environments/:environment_name/variables', {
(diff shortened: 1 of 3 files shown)


JakubBiel:
Hey @decyjphr. Can I ask for review please?

decyjphr (changes requested):
This will cause any settings that use the old way of declaring `custom branch policies` to break, right. But I think this is necessary to add support for tag type.

So the question is can we still support array of strings and assume those would be converts to `branch` type policy.

lib/plugins/environments.js
```diff
@@ -14,6 +14,16 @@ module.exports = class Environments extends Diffable {
...
           })
         }
+        // Expand custom_branch_policies to match GitHub schema.
+        if (environment.deployment_branch_policy && environment.deployment_branch_policy.custom_branch_policies) {
+          const policies = []
+          environment.deployment_branch_policy.custom_branch_policies.forEach(policy => {
+            policy.names.forEach(name => {
+              policies.push({ name: name, type: policy.type })
+            })
+          })
```
decyjphr:
```suggestion
  environment.deployment_branch_policy.custom_branch_policies.forEach(policy => {
    if (typeof policy === 'string') {
      // Convert string policy to object with type 'branch'
      policies.push({ name: policy, type: 'branch' })
    } else if (typeof policy === 'object' && Array.isArray(policy.names)) {
      policy.names.forEach(name => {
        policies.push({ name: name, type: policy.type })
      })
    }
  })
```

decyjphr (changes requested):

lib/plugins/environments.js
JakubBiel: Nice one. Is the indentation correct?

decyjphr: Sorry, I didn't check the indentation.

JakubBiel: Ok, I'll sort it. Should I also add tests for backward compatibility?

decyjphr: Sure, highly recommend using copilot to generate it for you to make your life easier 😄

JakubBiel: Added suggestion plus tests

decyjphr (approved):
Thanks!

martinjackson-itv:
@decyjphr this looks great, I'm keen to use the functionality. Do you know when this will get released? Thank you.

decyjphr:
@martinjackson-itv I am still in the middle of  testing all the PRs that have been merged and it's taking longer but I hope to release in a few days.

PR #811, opened 2025-04-14
Update variables in app.yml to actions_variables

a-magdy:


a-magdy:
Tried running the latest code and was getting this error when I use the manifest to create a new app:

Screenshot 2025-04-15 at 00 41 43

After few trials I figured that this changes doesn't cause any errors, haven't tested it fully yet to check that this covers the implementation of the existing setup

decyjphr (approved):
LGTM

Issue #780, opened 2025-03-05
creating new repo does not work as expected


maratsh:
## Problem Description
I'm trying to configure save-settings to create a new repo for me.
This is my config

```yaml
# save-settings/repos/test-please-ignore.yml

repository:
  # Create the repo if doesn't exist
  force_create: true
  name: test-please-ignore
  private: true
  organization: my-org
  description: "test please ignore"

```

```yaml
# save-settings/suborgs/test.yml
suborgrepos:
  - 'test-please-ignore'

rulesets:
  - name: Default
    target: branch
    enforcement: active

    conditions:
      ref_name:
        include:
          - ~DEFAULT_BRANCH
        exclude: []

    rules:
      - type: deletion
      - type: non_fast_forward # prevents force pushes

  - name: Pull Requests
    target: branch
    enforcement: active

    conditions:
      ref_name:
        include:
          - ~DEFAULT_BRANCH
        exclude: []

```

```yaml
# save-settings/deployment-settings.yml
restrictedRepos:
  include: ['^test-please-ignore$', 'rust-ci-test']
```

```yaml
# save-settings/settings.yml
repository:
  topics:
    - foobar
```

```yml
# workflow file
name: Safe Settings Sync
on:
#  schedule:
#    - cron: "0 */4 * * *"
  workflow_dispatch: {}
  push:
    branches:
      - main
jobs:
  safeSettingsSync:
    runs-on: ubuntu-latest
    environment: "main"
    env:
      # Version/tag of github/safe-settings repo to use:
      SAFE_SETTINGS_VERSION: 2.1.16

      # Path on GHA runner box where safe-settings code downloaded to:
      SAFE_SETTINGS_CODE_DIR: ${{ github.workspace }}/.safe-settings-code
    steps:
      # Self-checkout of 'admin' repo for access to safe-settings config:
      - uses: actions/checkout@v4

      # Checkout of safe-settings repo for running full sync:
      - uses: actions/checkout@v4
        with:
          repository: github/safe-settings
          ref: ${{ vars.SAFE_SETTINGS_VERSION || '2.1.16' }}
          path: ${{ github.workspace }}/.safe-settings-code
      - uses: actions/setup-node@v4
      - run: npm install
        working-directory: ${{ github.workspace }}/.safe-settings-code
      - run: npm run full-sync
        working-directory: ${{ github.workspace }}/.safe-settings-code
        env:
          GH_ORG: ${{ vars.SAFE_SETTINGS_GH_ORG }}
[...]

PR #833, opened 2025-05-23
fix: Re-enable creation of the new repositories

strainovic:
Hi all,

This PR fixes issue [safe-settings/issues/780](https://github.com/github/safe-settings/issues/780).

The code that returns the list of all repositories does not include one that does not exist.

Cheers! 👋


klutchell:
This change still requires `force_create: true`, correct? Just making sure that repo creation is still opt-in and not the default behaviour.

strainovic:
Hi @klutchell.

Yes, it requires `force_create: true`. It is here: https://github.com/github/safe-settings/blob/3e5d4fe23f2923e0d12e23d2c5d5473859776f3f/lib/plugins/repository.js#L121

Thanks

vaclav-ssvlabs:
It would be awesome to have this function fixed.

strainovic:
I'm not sure if I need to do something more to this PR before someone reviews and accepts it.

PR #857, opened 2025-07-02
feat: validate config with schema

marionebl:
This is an early implementation of json schema validation integrated with safe-settings directly. 

This will

1. Validate the `.github/settings.yml`
2. Validate sub org settings
3. Validate repo settings

The new validation is guarded by the env variable `VALIDATE_CONFIG_SCHEMA`.

I haven't manually tested this across all use cases yet but will do so over the next days.

For this PR I'm for now looking for feedback and an indication if the maintainers are interested in this.


marionebl:
@decyjphr is this something you're interested in as maintainer?

PR #858, opened 2025-07-03
fix: guard against result.action being undefined

marionebl:
I've observed the following recurring exception in handling `check_run` events. The stack trace indicates we're trying to read from `action` when it isn't defined. |

This accepts potentially partial check messages instead of crashing the event processing. 

```
    TypeError: Cannot read properties of undefined (reading 'additions')
       ~~~~~ redacted ~~~~~
        at Array.reduce (<anonymous>)
        at _Settings.handleResults ( ~~~~~ redacted ~~~~~)
```


marionebl:
@decyjphr could you have a look at this bugfix?

PR #859, opened 2025-07-07
fix: ignore merge commits for file diffs

marionebl:
Move from comparing the branch base against target to `pulls.listFiles`.

This solves for a bug where updating a given command with changes from
the target branch triggered the settings change detection.

Stacks on top of #858


marionebl:
@decyjphr could you have a look at this bug fix?

PR #888, opened 2025-10-03
Handle multiple config changes in a PR or Push event and process them as a batch

### LINKED ISSUE #887: Handle Push and Pull Request events having multiple config file changes.
## Situation:
The code was not handling cases where a push will have multiple file changes. It would only make the changes to the last changed file or cannot handle commits with multiple sub-org and repo files in it. Similarly a PR can have many file changes per commit and also may have many commits as part of the PR. All the changed files must be evaluated as part of the checks.

## Proposal
Rewrite the logic so that we get all the files in the PR and process all the repo config changes and sub-org config changes in a batch. Currently, the API we are using doesn't return all the files in the PR.
Similarly when there is a push to the default branch of the config repo, it could have multiple files in it and we should process all the changes as a batch.

> [!Note]
> The current implementation tried to handle it but was not working correctly

This will also make it easier for us to address issues like #818 where safe-settings deployed as an Actions workflow can also handle pull_request events and run the app in dry-run mode.

decyjphr:
This pull request refactors the way repository and sub-organization settings are synchronized, improving clarity and maintainability by consolidating logic and introducing a new method for handling multiple changes at once. The changes also update how file changes are detected in pull requests, aligning with GitHub's API best practices.

### DIFF
--- a/index.js
+++ b/index.js
@@ -266,8 +266,8 @@ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) =>
     let repoChanges = getAllChangedRepoConfigs(payload, context.repo().owner)
 
     let subOrgChanges = getAllChangedSubOrgConfigs(payload)
-    const dedupedRepos = [...new Set(repoChanges.map(r => r.repo))].map(name => {
-      return repoChanges.find(r => r.repo === name)
+    const dedupedRepos = [...new Set(repoChanges.map(r => r.repo))].map(repo => {
+      return repoChanges.find(r => r.repo === repo)
     })
     repoChanges = dedupedRepos
     const dedupedSubOrgs = [...new Set(subOrgChanges.map(s => s.repo))].map(repo => {
--- a/index.js
+++ b/index.js
@@ -263,9 +263,19 @@ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) =>
       return syncAllSettings(false, context)
     }
 
-    const repoChanges = getAllChangedRepoConfigs(payload, context.repo().owner)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -572,15 +571,10 @@ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) =>
     robot.log.debug(`Updating check run ${JSON.stringify(params)}`)
     await context.octokit.checks.update(params)
 
-    // guarding against null value from upstream libary that is
-    // causing a 404 and the check to stall
-    // from issue: https://github.com/github/safe-settings/issues/185#issuecomment-1075240374
-    if (check_suite.before === '0000000000000000000000000000000000000000') {
```
er0k, line 578: this seems like it would cause a regression of https://github.com/github/safe-settings/issues/185

why remove it?

decyjphr: Because we have changed the API call to a different one which does not require the before and after commits.

index.js
```diff
@@ -590,17 +589,10 @@ module.exports = (robot, { getRouter }, Settings = require('./lib/settings')) =>
     }
 
     const repoChanges = getChangedRepoConfigName(files, context.repo().owner)
-    if (repoChanges.length > 0) {
-      return Promise.all(repoChanges.map(repo => {
-        return syncSettings(true, context, repo, pull_request.head.ref)
-      }))
-    }
-
     const subOrgChanges = getChangedSubOrgConfigName(files)
```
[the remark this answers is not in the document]
decyjphr: This is correct, one method returns configs which are objects, the other returns name objects.

PR #919, opened 2026-01-09
Change helm defaults to be more newcomer friendly

### LINKED ISSUE #918: Helm chart deploy defaults could be better
## New Feature

The current Helm chart has some issues that make it difficult for newcomers to install.  

1. When ingress.create is true and service.port is not the default 3000, the containerPort in the deployment is incorrectly set to the service port.  This should always be 3000 because that's what the application exposes itself as inside the container.  
2. Required environment variables are not present as examples (e.g. APP_ID, PRIVATE_KEY, and WEBHOOK_SECRET)
3. Example validators are present and not commented out, so if a user doesn't want those validators, they have to explicitly set them to null instead of omit them.  
4. Min replica counts default to 2, but I would argue that 1 would be a more cost effective default.

tylerohlsen:
Fixes #918


tylerohlsen:
@decyjphr sorry for the ping, but would you be able to look this over this change?

PR #926, opened 2026-02-02
fix(variables): add noop mode support

### LINKED ISSUE #925: The variables plugin does NOT implement noop support

The `variables` plugin (added in #819) does not respect the `nop` (no-operation/dry-run) flag. When running safe-settings in noop mode, the variables plugin still makes actual API calls to create, update, and delete repository variables.

Affected methods in `lib/plugins/variables.js`:
- `add()` (line 156) - calls `POST /repos/:org/:repo/actions/variables` without noop check
- `update()` (line 87) - calls `PATCH`, `POST`, and `DELETE` endpoints without noop check
- `remove()` (line 181) - calls `DELETE /repos/:org/:repo/actions/variables/:variable_name` without noop check

When running in noop mode (`nop: true`), the variables plugin should:
1. NOT make any actual GitHub API calls
2. Return `NopCommand` objects describing what changes *would* be made
3. Match the behavior of other plugins like `environments.js` which use `nopifyRequest()` or inline noop checks

```
No error - the plugin silently makes real changes even in dry-run mode
```

Running our own self-hosted instance as a GitHub App

GitHub.com

Snapshot from `main-enterprise` branch (Dec 4, 2025)

N/A - using GitHub.com

klutchell:
Fixes: https://github.com/github/safe-settings/issues/925

- Add noop mode support to Variables plugin add/remove/update methods
- Return NopCommand instead of making API calls when nop=true
- Add comprehensive tests for noop mode behavior

Refactor Variables plugin to match the single-item Diffable contract
used by labels, milestones, and other plugins. The previous update()
reimplemented sync() logic internally; now each method handles one item
and lets Diffable.sync() orchestrate iteration.

- Simplify update() from 90-line array-diffing to single-item PATCH
- Simplify changed() from JSON.stringify comparison to value check
- Remove getChanged(), lodash dependency, .then(res=>res) no-ops
- Match labels.js nop return pattern: Promise.resolve([NopCommand])
- Fix inconsistent toUpperCase() between add/remove/update
- Let errors propagate to Diffable.sync() instead of swallowing
- Normalize find() to strip API metadata fields (created_at, etc.)

### DIFF
--- a/lib/plugins/variables.js
+++ b/lib/plugins/variables.js
@@ -1,4 +1,3 @@
-const _ = require('lodash')
 const Diffable = require('./diffable')
 const NopCommand = require('../nopcommand')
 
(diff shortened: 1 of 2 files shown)


klutchell:
@decyjphr this one should be good to go!

klutchell:
@decyjphr conflicts resolved

magnusp:
Just to add some additional urgency to getting this released: currently on latest 2.1.21 it allows unconditional overwriting of variables on the organization level when a PR is opened. Running safe-settings in a workflow could mitigate this by stripping the `variables` section from configuration files before starting safe-settings, but that on the other hand completly removes management of variables.

decyjphr (approved):
Good to 🚢

klutchell:
Thanks @decyjphr !

PR #943, opened 2026-03-05
fix: prevent race conditions by creating new objects for repository configurations

John15321:
## Summary

Fixes a race condition where `repository.name` could leak between repos during concurrent processing, and prevents `name` from appearing in dry-run diffs when not explicitly configured.

## Problem

1. **Race condition**: `Object.assign(repoConfig, {...})` mutates the shared `this.config.repository` object. When `updateRepos()` processes multiple repos concurrently via `Promise.all`, one repo's name can overwrite another's between async yields.

2. **Spurious diffs**: `name` always appears in dry-run diffs because Safe-Settings injects it internally but it's not in `ignorableFields`.

## Changes

- `lib/settings.js`: Use `Object.assign({}, repoConfig, {...})` to create new objects instead of mutating shared references
- `lib/plugins/repository.js`: Add `'name'` to `ignorableFields` to hide it from diffs unless explicitly configured for renaming

## Testing

All 112 unit tests pass.

Related to https://github.com/github/safe-settings/issues/944

### DIFF
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -7,7 +7,6 @@ const ignorableFields = [
   'id',
   'node_id',
   'full_name',
-  'name',
   'private',
   'fork',
   'created_at',
--- a/lib/plugins/repository.js
+++ b/lib/plugins/repository.js
@@ -7,6 +7,7 @@ const ignorableFields = [
   'id',
   'node_id',
   'full_name',
+  'name',
   'private',
   'fork',
   'created_at',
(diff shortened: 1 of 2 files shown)


John15321:
Just fyi, I am not a JS developer at all, so this code analysis and fix was aided by AI, therefore please approach this code with skepticism, but the actual problem we are facing when trying to use `safe-settings` remains

John15321:
Hi @decyjphr could you please take a look at this PR?

PR #988, opened 2026-06-01
Skip settings sync for repos that are already archived

pcrumley:
My deployment of github safe-settings fails at the sync step on archived repos, because GitHub treats archived repos as read-only and rejects any settings update (repository, branch protection, labels, teams, etc.) with a 403 "Repository was archived so is read-only." error.

`updateRepos` only used the archive plugin's state to decide whether to archive/unarchive a repo, but it still ran the repository plugin and all child plugins unconditionally. For repos that are already archived and are staying archived, this issues a `repos.update()` (and other writes) against a read-only repo every sync, which 403s, gets logged as an error and surfaces as a failing check run.

Skip the repository and child plugins when a repo is archived and is not being unarchived in this run. Newly-archived repos (shouldArchive) are still configured first and archived last, and repos being unarchived are configured after the unarchive, so those paths are unchanged.

Add regression tests covering already-archived, newly-archived and non-archived repos.

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -370,7 +370,7 @@ ${this.results.reduce((x, y) => {
         const RepoPlugin = Settings.PLUGINS.repository
 
         const archivePlugin = new Archive(this.nop, this.github, repo, repoConfig, this.log)
-        const { shouldArchive, shouldUnarchive } = await archivePlugin.getState()
+        const { isArchived, shouldArchive, shouldUnarchive } = await archivePlugin.getState()
 
         if (shouldUnarchive) {
           this.log.debug(`Unarchiving repo ${repo.repo}`)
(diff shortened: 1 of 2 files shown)


pcrumley:
The issue this is trying to fix is that if you have set a repo to be archived then someone changes some of the settings manually by click-opsing them, the safe-settings bot will try to patch those settings back to the settings in the repo. However it is unable to do that because github api 403's whenever you make a request to change a setting of a archived repo.

I can provide a minimal example if you would like. I also am not sure if this is the best way to solve the problem. The other solution would be to unarchive the repo, set all the settings back to what was in the config, then re-archive the repo.

decyjphr:
Closing this as #991 already resolved it. Please reopen if that is not the case.

PR #997, opened 2026-06-15
fix: prevent phantom team diffs caused by slug vs name mismatch

fgenois-coveo:
## Problem

When running safe-settings in NOP mode (or any diff-reporting path), teams are incorrectly reported as additions + deletions even when nothing has changed.

**Root cause:** `MergeDeep.processArrays()` resolves item identity via `NAME_FIELDS`, but `slug` was missing from that list. For teams, the GitHub API returns both `name` (display name like "Dev Tooling") and `slug` (like "dev-tooling"). YAML configs use the slug value in the `name` field. Since the API object's `name` differs from the YAML's `name`, MergeDeep treats them as different items.

The sync path in `teams.js` already handles this correctly (`existing.slug === attrs.name.toLowerCase()`), so the mismatch only affects diff reporting.

## Fix

1. Add `slug` (first position) to `NAME_FIELDS` — API team objects now resolve identity via `slug`, while YAML objects (which lack `slug`) resolve via `name`. Both produce the same value → correct match.

2. Add `title` to `NAME_FIELDS` — milestones use `title` as identity but it was missing from the list.

3. Remove the hardcoded `if (item.slug) return item.slug` special case — the priority-ordered `NAME_FIELDS` handles this uniformly.

4. Guard `compareDeepIfVisited` against mixed primitive/object arrays — prevents crashes when branch protection compares `users: ["test"]` against `users: [{login: "test", type: "User"}]`.

5. Strip identity fields per-item — when source uses `name` and target uses `slug`, each side strips its own identity field before deep comparison.

**Backward compatible** — no other plugin returns objects with a `slug` field, and `title` is only used by milestones (which have no `name` field).

## Test matrix

| Source (YAML) | Target (API response) | Expected | Before | After |
|---|---|---|---|---|
| `{name: "dev-tooling", permission: "admin"}` | `{name: "Dev Tooling", slug: "dev-tooling", permission: "admin"}` | No diff | ❌ Add+Delete | ✅ No diff |
| `{name: "dev-tooling", permission: "push"}` | `{name: "Dev Tooling", slug: "dev-tooling", permission: "admin"}` | Modification | ❌ Add+Delete | ✅ Modification |
| `{name: "My Team", permission: "admin"}` | `{name: "My Team", slug: "my-team", permission: "admin"}` | No diff | ✅ No diff | ✅ No diff |
| `{name: "new-team", permission: "push"}` | (empty) | Addition | ✅ | ✅ |
| (empty) | `{name: "Old Team", slug: "old-team"}` | Deletion | ✅ | ✅ |

### DIFF
--- a/lib/mergeDeep.js
+++ b/lib/mergeDeep.js
@@ -184,7 +184,7 @@ class MergeDeep {
     if (source.length < target.length) {
       const dels = target.filter(item => {
         if (this.isObjectNotArray(item)) {
-          return !source.some(sourceItem => GET_NAME_USERNAME_PROPERTY(item) === GET_NAME_USERNAME_PROPERTY(sourceItem))
+          return !source.some(sourceItem => { const s = GET_NAME_USERNAME_PROPERTY(sourceItem); const t = GET_NAME_USERNAME_PROPERTY(item); if (s === t) return true; if (item.slug && typeof s === "string") return item.slug === s || item.slug === s.toLowerCase(); return false; })
         } else {
           return !source.includes(item)
         }
@@ -222,7 +222,7 @@ class MergeDeep {
       // Elements that are not in target are additions
       additions[key] = combined.filter(item => {
         if (this.isObjectNotArray(item)) {
-          return !target.some(targetItem => GET_NAME_USERNAME_PROPERTY(item) === GET_NAME_USERNAME_PROPERTY(targetItem))
+          return !target.some(targetItem => { const s = GET_NAME_USERNAME_PROPERTY(item); const t = GET_NAME_USERNAME_PROPERTY(targetItem); if (s === t) return true; if (targetItem.slug && typeof s === "string") return targetItem.slug === s || targetItem.slug === s.toLowerCase(); return false; })
         } else {
           return !target.includes(item)
         }
@@ -233,7 +233,7 @@ class MergeDeep {
       // Elements that not in source are deletions
       deletions[key] = combined.filter(item => {
         if (this.isObjectNotArray(item)) {
-          return !source.some(sourceItem => GET_NAME_USERNAME_PROPERTY(item) === GET_NAME_USERNAME_PROPERTY(sourceItem))
+          return !source.some(sourceItem => { const s = GET_NAME_USERNAME_PROPERTY(sourceItem); const t = GET_NAME_USERNAME_PROPERTY(item); if (s === t) return true; if (item.slug && typeof s === "string") return item.slug === s || item.slug === s.toLowerCase(); return false; })
         } else {
           return !source.includes(item)
         }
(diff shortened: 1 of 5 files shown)


lib/mergeDeep.js
```diff
@@ -1,9 +1,13 @@
 const mergeBy = require('./mergeArrayBy')
 const DeploymentConfig = require('./deploymentConfig')
 
-const NAME_FIELDS = ['name', 'username', 'actor_id', 'login', 'type', 'key_prefix', 'context']
+const NAME_FIELDS = ['slug', 'name', 'username', 'actor_id', 'login', 'type', 'key_prefix', 'context', 'title']
 const NAME_USERNAME_PROPERTY = item => NAME_FIELDS.find(prop => Object.prototype.hasOwnProperty.call(item, prop))
-const GET_NAME_USERNAME_PROPERTY = item => { if (NAME_USERNAME_PROPERTY(item)) return item[NAME_USERNAME_PROPERTY(item)] }
+const GET_NAME_USERNAME_PROPERTY = item => {
```
fgenois-coveo: why does this need changing? try to change as little code as possible

fgenois-coveo: AI: You're right — the refactor was unnecessary. The original one-liner already handles the falsy case (returns undefined implicitly). Reverted to the original form; the only change in this block is now the addition of slug and title to NAME_FIELDS.

lib/mergeDeep.js
```diff
@@ -248,8 +248,13 @@ class MergeDeep {
       modifications.push({})
       additions.push({})
       deletions.push({})
-      if (visited[id]) {
```
fgenois-coveo, line 251: Explain why this change is necessary. and remove change if unnecessary; this seems like an ugly fix

fgenois-coveo: AI: The guard is necessary because `visited[id]` can be a primitive string — in `processArrays`, when items are not objects, they get stored as `visited[a] = a` (a string). Later, an object with the same identity key (e.g. `{name: "foo"}`) hits `compareDeepIfVisited` and `visited[id]` resolves to that primitive string. Without the guard, `compareDeep` would be called on a string, which breaks.

Simplified: removed the redundant `isObjectNotArray(a)` half since the caller already guarantees `a` is an object (line 196: `if (this.isObjectNotArray(a))`). Only `isObjectNotArray(visited[id])` is needed. Added a comment explaining the edge case.

fgenois-coveo: eh, why is that required for our change? is that a fix for another problem? gimme a full example of data and sequence.

fgenois-coveo: AI: You're right — the guard was unnecessary. `compareDeepIfVisited` is only called from within the `if (this.isObjectNotArray(a))` branch, so `a` is always an object. And `visited[id]` is only populated from that same branch (`visited[id] = a`), so it's always an object too. The primitive branch uses the primitive itself as key (`visited[a] = a`), which can't collide with `GET_NAME_USERNAME_PROPERTY()` results from objects.

Removed both `isObjectNotArray` guards, restoring the original `if (visited[id])` truthy checks. The identity-field stripping before `compareDeep` is kept — that IS needed for our change because adding `slug` to `NAME_FIELDS` means matched objects may have different identity properties (one has `slug`, the other `name`), which would otherwise appear as a spurious diff.

PR #1076, opened 2026-09-18
Add opt-in PR comment deduplication by minimizing outdated comments.

TovaBecker:
Every push to a PR posts a brand-new safe-settings comment, so PRs with several pushes end up with a growing pile of near-identical diffs, making it hard to tell what's current. This collapses the outdated comment (via GitHubs "Hide comment" mechanism) before posting the new one, so only the latest is visible while keeping history intact for anyone who wants to expand it. 

Gated behind PR_COMMENT_DEDUPE_ENABLED (default false) so existing deployments see no change unless they opt in.

This requires the Github App to have `issues: write` permission(which it already has by default as of today)

image


TovaBecker:
Quick clarification on the `README.md:548` finding from Copilot's latest review: that's flagging a pre-existing naming inconsistency in the repo (the docs already said `ENABLE_PR_COMMENT` while `lib/env.js` has always read `CREATE_PR_COMMENT`), not something introduced by this PR's `PR_COMMENT_DEDUPE_ENABLED` code. Fixed it anyway in `99c883f` since it's a one-line doc correction and we were already touching that section, but wanted to flag that it's unrelated to the dedupe feature itself.

TovaBecker:
Closed this one for a new fresh pr.

PR #1079, opened 2026-09-18
Add opt-in PR comment deduplication by minimizing outdated comments

TovaBecker:
## Summary

Every push to a PR posts a brand-new safe-settings comment, so PRs with several pushes end up with a growing pile of near-identical diffs, making it hard to tell what's current. This collapses the outdated comment (via GitHub's "Hide comment" mechanism) after posting the new one, so only the latest is visible while keeping history intact for anyone who wants to expand it.

Gated behind `PR_COMMENT_DEDUPE_ENABLED` (default `false`) so existing deployments see no change unless they opt in. 

This requires the Github App to have `issues: write` permission(which it already has by default as of today)

image

Second commit  is an isolated, pre-existing doc fix (`ENABLE_PR_COMMENT` → `CREATE_PR_COMMENT`), unrelated to this feature.


TovaBecker:
This replaces #1076 (closed) — same code, squashed into a clean two-commit history (feature + an unrelated pre-existing naming fix). Continuing review-driven fixes here.

TovaBecker:
Re: the GraphQL-plugin concern on `lib/settings.js` — I checked this empirically and I don't think it holds. `this.github.graphql` isn't a plugin at all in current Octokit versions; `@octokit/core`'s own constructor sets it unconditionally (`this.graphql = withCustomRequest(this.request).defaults(...)`, [source](https://github.com/octokit/core.js/blob/main/src/index.ts)), independent of the plugin list. I instantiated the exact plugin stack from `lib/proxyAwareProbotOctokit.js` (`throttling, retry, paginateRest, legacyRestEndpointMethods`) directly and confirmed `typeof instance.graphql === 'function'`.

While verifying that, though, I did find a real, separate, pre-existing issue: `lib/proxyAwareProbotOctokit.js` requires `@octokit/plugin-enterprise-compatibility` via CommonJS `require()`, but the currently-installed version (6.0.3) only declares an `"import"` condition in its `package.json` `exports` map — no `"require"` — so that `require()` call throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. That means the AWS Lambda entry point (`handler.js`, the only place that loads this file) currently can't even start up. Unrelated to this PR's dedupe feature — happy to file a separate issue for it if useful.

PR #1082, opened 2026-09-23
fix(environments): stop mutating the shared config when expanding branch policies

### LINKED ISSUE #876: Environment creation doesn't set custom branch policies
## Problem Description

Currently the environment configuration doesn't properly create environments with custom branch policies. It creates the environment and sets it to use "Custom branch policies" but it doesn't create the rules for the selected branches/tags.

We are using the following configuration to set our environment but it's not applying the branch policies:

```
environments:
  - name: prod
    prevent_self_review: true
    deployment_branch_policy:
      protected_branches: false
      custom_branch_policies: 
         - names: ["main"]
            type: "branch"
```

Image

We also tried the legacy format, org level and suborg level; no success. 

We know the environment creation and configuration it's a two step operation from the API side (one call to create the environment and a second one to create the branch protection policy). It looks like it's only doing the first one. 

I would expect to have the environment created and the defined custom branch policy set.

### Safe-settings version
Commit SHA: 861d70f1d24427f523363a83709429d7aa041510

ethanejones:
The `Environments` constructor rewrites `custom_branch_policies` on the config objects it is given. Suborg config is one object shared by every repo in the suborg, so the first repo turns `['main']` into `[{ name: 'main', type: 'branch' }]` and every later repo sees objects that match neither expansion branch. Those repos end up with custom branch policies switched on and an empty allowlist, which blocks every deployment with no error.

This clones the entries before normalising them. Two tests cover it: the caller's config is left untouched, and a second plugin built from the same config gets the same expanded policies as the first.

The same two tests against `main-enterprise` (left) and this branch (right):

jest output before and after: both tests fail upstream with an empty allowlist on the second repo, both pass with the fix

Fixes #876. The symptom there (custom policies enabled, no rules created, at org and suborg level) matches what we saw on an org with four suborgs declaring the policy.


ethanejones:
@decyjphr when you next do a pass at reviews on this repo this PR is small. 🙂
Just 1 line in `environments.js` plus two tests and it'll close #876.
Thanks!

